PR #124, opened 2016-06-11
google-auth-client: check for empty or null userId

mgenov:
When userId is not specified no requests need to be send to backed, cause it's obvious that user is unknown.


lesv:
Restarted travis

lesv:
@garrettjonesgoogle Approving and merging.

PR #135, opened 2016-08-22
Implement customizable landing pages after login / use "127.0.0.1" instead of "localhost"

### LINKED ISSUE #128: Customizable login success / failure landing pages for LocalServerReceiver
`LocalSeverReceiver` has [a hard-coded HTML response](https://github.com/google/google-oauth-java-client/blob/dev/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java#L232) after successful login.

It would be nice if clients can configure and customize success/failure landing pages.

chanseokoh:
Fixes #128.

The change in `LocalServerReceiver` is actually very small. Added several unit tests, which verify actual responses of the local server.

PTAL.

### DIFF
--- a/google-oauth-client-jetty/pom.xml
+++ b/google-oauth-client-jetty/pom.xml
@@ -76,5 +76,10 @@
       <groupId>org.mortbay.jetty</groupId>
       <artifactId>jetty</artifactId>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/google-oauth-client-jetty/pom.xml
+++ b/google-oauth-client-jetty/pom.xml
@@ -79,6 +79,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <version>4.8.2</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
+++ b/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
@@ -45,6 +45,8 @@ import javax.servlet.http.HttpServletResponse;
  */
 public final class LocalServerReceiver implements VerificationCodeReceiver {
 
+  private static final String LOCALHOST = "127.0.0.1";
+
   private static final String CALLBACK_PATH = "/Callback";
 
   /** Server or {@code null} before {@link #getRedirectUri()}. */
@@ -88,7 +90,7 @@ public final class LocalServerReceiver implements VerificationCodeReceiver {
    * </p>
    */
   public LocalServerReceiver() {
-    this("127.0.0.1", -1);
+    this(LOCALHOST, -1);
   }
 
   public LocalServerReceiver(String successLandingPageUrl, String failureLandingPageUrl) {
--- a/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
+++ b/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
@@ -69,6 +69,18 @@ public final class LocalServerReceiver implements VerificationCodeReceiver {
   private final String host;
 
   /**
+   * URL to an HTML page to be shown (via redirect) after successful login. If null, a canned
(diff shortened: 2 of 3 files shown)


chanseokoh:
For some unknown reason, my tests are failing on Travis, so I'll take a look and fix them.

chanseokoh:
I've fixed tests. PTAL.

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
```diff
@@ -69,14 +69,32 @@
...
+  private String successLandingPageUrl;
+
+  /**
+   * URL to an HTML page to be shown (via redirect) after failed login. If null, a canned
+   * default landing page will be shown (via direct response).
+   */
+  private String failureLandingPageUrl;
+
+  /**
+   * Constructor that starts the server on {@code "127.0.0.1"} selects an unused port.
```
elharo: selects --> and

chanseokoh:
Ping. PTAL.

anthmgoogle:
@ejona86 @garrettjonesgoogle and @anthmgoogle should review this.

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
```diff
@@ -69,14 +69,32 @@
...
+  /**
+   * Constructor that starts the server on {@code "127.0.0.1"} and an unused port.
    *
    * <p>
    * Use {@link Builder} if you need to specify any of the optional parameters.
    * </p>
    */
   public LocalServerReceiver() {
-    this("localhost", -1);
+    this("127.0.0.1", -1);
```
anthmgoogle: Use a constant for the IP address.

anthmgoogle:
LGTM. A non-blocking suggestion.

google-oauth-client-jetty/pom.xml
```diff
@@ -76,5 +76,10 @@
       <groupId>org.mortbay.jetty</groupId>
       <artifactId>jetty</artifactId>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
```
garrettjonesgoogle: We should specify versions explicitly, or builds won't be repeatable.

chanseokoh: I put `<version>4.8.2</version>` as I can see in the parent `pom.xml`.

garrettjonesgoogle:
LGTM

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
```diff
@@ -45,6 +45,8 @@
  */
 public final class LocalServerReceiver implements VerificationCodeReceiver {
 
+  private static final String LOCALHOST = "127.0.0.1";
```
ejona86: I see [the referenced doc](https://developers.google.com/identity/protocols/OAuth2InstalledApp), but it also says:

>  You should query your platform for the relevant loopback IP, and start an HTTP listener on a random available port

In Java, that would be done via `InetAddress.getLoopbackAddress()`. Although I also notice that other parts of the code will break with IPv6. Although in the same document it suggests doing `http://::1:port`, which would be broken in the same way, as I would have expected `http://[::1]:port`.

Looking at the history, before July the document suggested to use `localhost`, but it was changed to the text we see "as it plays better with firewalls such as the default Windows 10 firewall".

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
```diff
@@ -69,14 +71,32 @@
...
    * <p>
    * Use {@link Builder} if you need to specify any of the optional parameters.
    * </p>
    */
   public LocalServerReceiver() {
-    this("localhost", -1);
+    this(LOCALHOST, -1);
+  }
+
+  public LocalServerReceiver(String successLandingPageUrl, String failureLandingPageUrl) {
```
ejona86: Instead of having another public constructor, it looks like these options should be added to the Builder instead.

ejona86:
After the configuration gets moved to the builder, this would LGTM and I could merge.

chanseokoh:
The landing page configuration has been moved to the builder. PTAL.

ejona86:
LGTM. Cherry-picking

PR #136, opened 2016-09-06
Cancelable waitForCode() / port assignment fix

### LINKED ISSUE #126: Choosing unused local port in LocalServerReceiver is flawed
There is no guarantee that the unused port returned [here](https://github.com/google/google-oauth-java-client/blob/dev/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java#L96) will remain unused until the Jetty server starts. Any process can intercept the port before the server actually starts, and this will lead to an error when starting the server.

A more correct and also much simpler way is to let Jetty choose an unused port and query later which port it actually picked up:

http://stackoverflow.com/questions/8884865/how-to-discover-embedded-jetty-port-after-requesting-random-available-port

### LINKED ISSUE #127: Mechanism to cancel waitForCode in LocalServerReceiver
Currently, callers of `waitForCode()` are blocked and cannot escape the method until the server handles the login redirect. In IDEs, this is a critical problem, as users cannot cancel a login flow. Because of this, [Google Cloud Tools for IntelliJ](https://github.com/GoogleCloudPlatform/gcloud-intellij) already [duplicated the code](https://github.com/GoogleCloudPlatform/gcloud-intellij/blob/master/google-account-plugin/src/com/google/cloud/tools/intellij/login/CancellableServerReceiver.java) of `LocalServerReceiver`. I may also duplicate the code soon for [Google Cloud Tools for Eclipse](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools).

One idea would be to make `stop()` (or a new method) escape a caller that is blocked or will be blocked on `waitForCode()`.

Another idea would be to make `waitForCode()` accept a timeout so that callers can poll.

chanseokoh:
Fixes #127 #126

- `stop()` will now release any caller that is currently blocked or will be otherwise blocked on `waitForCode()`. To ensure this is working as intended, I've added a number of extensive tests.

- Jetty itself will now pick an unused port. Fixes #126, and the code has been simplified consequently.

### DIFF
--- a/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
+++ b/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
@@ -45,6 +45,8 @@ import javax.servlet.http.HttpServletResponse;
  */
 public final class LocalServerReceiver implements VerificationCodeReceiver {
 
+  private static final String LOCALHOST = "127.0.0.1";
+
   private static final String CALLBACK_PATH = "/Callback";
 
   /** Server or {@code null} before {@link #getRedirectUri()}. */
@@ -88,7 +90,7 @@ public final class LocalServerReceiver implements VerificationCodeReceiver {
    * </p>
    */
   public LocalServerReceiver() {
-    this("127.0.0.1", -1);
+    this(LOCALHOST, -1);
   }
 
   public LocalServerReceiver(String successLandingPageUrl, String failureLandingPageUrl) {
--- a/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
+++ b/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
@@ -69,6 +69,18 @@ public final class LocalServerReceiver implements VerificationCodeReceiver {
   private final String host;
 
   /**
+   * URL to an HTML page to be shown (via redirect) after successful login. If null, a canned
+   * default landing page will be shown (via direct response).
+   */
+  private String successLandingPageUrl;
+
+  /**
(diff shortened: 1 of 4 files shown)


chanseokoh:
I am not sure why this PR is still showing changes that have been already merged in #135.

@ejona86 @garrettjonesgoogle @anthmgoogle PTAL.

chanseokoh:
Ah, I didn't sync my fork with the upstream. Now everything looks good. PTAL.

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
```diff
@@ -128,22 +119,16 @@ public String getRedirectUri() throws IOException {
...
-        gotAuthorizationResponse.awaitUninterruptibly();
-      }
-      if (error != null) {
-        throw new IOException("User authorization failed (" + error + ")");
-      }
-      return code;
-    } finally {
-      lock.unlock();
+    waitUnlessSignaled.acquireUninterruptibly();
+    if (error != null) {
```
ejona86: It looks like this won't fail throw if `stop()` is called. So `waitForCode()` will return `null`. I don't really know if that's good or bad, but it should probably be in some JavaDoc.

ejona86:
This looks great to me (LGTM). One minor comment.

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
chanseokoh: I think returning null makes sense for `stop()`: `error` comes from HTTP request parameter, and stopping a server doesn't really have to throw an IOException. I'll add a JavaDoc for `waitForCode()` noting that `null` will be returned if `stop()` is called.

anthmgoogle:
LGTM.

PR #138, opened 2016-09-20
Clarify difference between ExpiresIn and ExpirationTime

### LINKED ISSUE #137: Clarify difference between Credential's ExpirationTimeMilliseconds vs ExpiresInSeconds
The difference between `[gs]etExpiresInSeconds()` and  `[gs]etExpirationTimeMilliseconds()` is subtle and should be called out explicitly in the javadocs.  One is relative to now, whereas the other is fixed (ok, relative to the Java epoch).  It is confusing to developers, and has resulted in developers manually re-anchoring expiration times by calculating `System.currentTImeMills() + credential.getExpiresInSeconds() * 1000`.

For example, [Credential#getExpirationTimeMilliseconds()](https://github.com/google/google-oauth-java-client/blob/master/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/Credential.java#L392) says:

```java
/** Expected expiration time in milliseconds or {@code null} for none. */
  public final Long getExpirationTimeMilliseconds() {
```

That would be better explicitly referencing the epoch.

```java
/**
  * Expected expiration time in milliseconds relative to the 
  * {@link System#currentTimeMills() Java epoch}  or {@code null} for none.
  */
  public final Long getExpirationTimeMilliseconds() {
```

Unfortunately this javadoc to have spread quite far.

briandealwis:
Clarify the differences in where the times are anchored.

Fixes #137

### DIFF
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/Credential.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/Credential.java
@@ -389,7 +389,10 @@ public class Credential
     return this;
   }
 
-  /** Expected expiration time in milliseconds or {@code null} for none. */
+  /**
+   * Expected expiration time in milliseconds relative to the
+   * {@link System#currentTimeMillis() Java epoch}, or {@code null} for none.
+   */
   public final Long getExpirationTimeMilliseconds() {
     lock.lock();
     try {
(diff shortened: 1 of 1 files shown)


chanseokoh:
@anthmgoogle @ejona86 @garrettjonesgoogle ptal.

PR #139, opened 2016-09-22
Add OSGi Metadata

### LINKED ISSUE #125: Add OSGI metadata to the JAR
The Cloud Tools for Eclipse team https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools would like to add OSGI metadata to the final jar file. This will enable us to pull in this dependency from Maven Central rather than copying the binary file into our source tree. This will also assist any other developer who wants to reference this library from an Eclipse plug-in or another OSGI application. 

This should have minimal impact on anyone not using OSGI. That is, it's just some extra entries in MANIFEST.MF everyone else will ignore. E.g. 

Bundle-ManifestVersion: 2
Bundle-SymbolicName: com.google.oauth-client
Bundle-Version: 1.2.3
Export-Package: com.google.oauth-client;version="1.2.3";uses:="com.google.httpclient"
Import-Package: com.google.httpclient;version="[1.0, 3)"

(That's just an example, not an actual working patch.)
See http://blog.osgi.org/2013/01/get-help-adding-osgi-metadata-to-your.html

If you're amenable to this, I'll send a real patch.

chanseokoh:
Fixes #125.

Adding metadata to the following modules:
`google-oauth-client`
`google-oauth-client-java6`
`google-oauth-client-servlet`
`google-oauth-client-appengine`
`google-oauth-client-jetty`

I skipped the following module, as it seems to be a special module for assembling a zip:
`google-oauth-client-assembly`

### DIFF
--- a/google-oauth-client-appengine/pom.xml
+++ b/google-oauth-client-appengine/pom.xml
@@ -36,7 +36,6 @@
         <executions>
           <execution>
             <id>jar</id>
-            <phase>compile</phase>
             <goals>
               <goal>jar</goal>
             </goals>
@@ -48,7 +47,6 @@
         <executions>
           <execution>
             <id>source-jar</id>
-            <phase>compile</phase>
             <goals>
               <goal>jar</goal>
             </goals>
--- a/google-oauth-client-appengine/pom.xml
+++ b/google-oauth-client-appengine/pom.xml
@@ -24,26 +24,6 @@
         </configuration>
       </plugin>
       <plugin>
-        <groupId>org.apache.felix</groupId>
-        <artifactId>maven-bundle-plugin</artifactId>
-        <version>3.2.0</version>
-        <configuration>
-          <instructions>
-            <Bundle-SymbolicName>com.google.oauth-client-appengine</Bundle-SymbolicName>
-            <Bundle-RequiredExecutionEnvironment>JavaSE-1.6</Bundle-RequiredExecutionEnvironment>
-          </instructions>
-        </configuration>
-        <executions>
-          <execution>
-            <id>bundle-manifest</id>
-            <phase>compile</phase>
-            <goals>
-              <goal>manifest</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
         <artifactId>maven-jar-plugin</artifactId>
         <configuration>
           <archive>
@@ -56,7 +36,6 @@
         <executions>
           <execution>
             <id>jar</id>
-            <phase>compile</phase>
             <goals>
               <goal>jar</goal>
             </goals>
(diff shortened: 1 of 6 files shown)


chanseokoh:
@briandealwis

- I set the execution phase of `org.apache.felix` to `compile` instead of `process-classes` to maintain the same execution order: `org.apache.felix` (`compile`) --> `maven-jar-plugin` (`compile`) --> `maven-datanucleus-plugin` (`process-classes`).
- I set `Bundle-RequiredExecutionEnvironment` to `JavaSE-1.6`, as I see there is `google-oauth-client-java6`.

Do they look good?

chanseokoh:
After investigation, I learned that `mvn package` will anyway re-package and re-write jars in the `package` phase that were created in the `compile` phase. Therefore, I removed all the `<phase>compile</phase>` setting for `maven-jar-plugin`, as I see no point of packaging twice (unless if it's for convenience for development).

google-oauth-client-appengine/pom.xml
```diff
@@ -30,12 +30,12 @@
             <manifest>
               <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
             </manifest>
+            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
           </archive>
         </configuration>
         <executions>
```
briandealwis: You can drop these `maven-jar-plugin` `<executions>` specs entirely since they just duplicate what's already specified by the default lifecycle.

chanseokoh: Done. What was happening before was that this `<execution>` had a different `<id>jar</id>` than the default `<id>default-jar</id>` (along with `<phase>compile</phase>`), so that's why it was packaging jars twice.

chanseokoh:
@anthmgoogle @ejona86 @garrettjonesgoogle PTAL

anthmgoogle:
LGTM although I don't have much context on this build configuration stuff. We should wait on a review by both @ejona86 and @garrettjonesgoogle, and potentially at-wonderfly if we still feel short on context.

garrettjonesgoogle:
LGTM, although like @anthmgoogle , I haven't worked with this packaging stuff much so I don't have any insight into it.

chanseokoh:
ping @ejona86

chanseokoh:
ping @ejona86

chanseokoh:
ping @ejona86

ejona86:
LGTM. Like the others, I don't think I would really be able to spot a bug, but it does appear sensible.

chanseokoh:
@anthmgoogle could you merge this?

elharo:
ping @anthmgoogle Can you please merge this?

PR #144, opened 2017-02-04
Allow credential without explicit expire time

galacticvac:
Certain sites provide an access token without an expiration. Currently this causes a NullPointerException. This change is to allow valid credentials that don't have an expiry.


galacticvac:
I signed it!

tswast:
@garrettjonesgoogle PTAL (or is someone else maintaining this?)

PR #155, opened 2017-05-23
Migrate google-oauth-client-jetty to Jetty 8.

evantill:
use `org.eclipse.jetty` instead of `org.mortbay.jetty`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 target/
 bin/
+.idea
+*.iml
--- a/google-oauth-client-jetty/pom.xml
+++ b/google-oauth-client-jetty/pom.xml
@@ -84,8 +84,8 @@
       <artifactId>google-oauth-client-java6</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.mortbay.jetty</groupId>
-      <artifactId>jetty</artifactId>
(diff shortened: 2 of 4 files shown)


evantill:
I signed the CLA.

chanseokoh:
@ejona86

lesv:
Be careful with this, think about GAE J7 vs J8.  This should be good on J8, but not J7.

evantill:
@lesv could you explain a bit more please ? What can I do to help on this ?

lesv:
@evantill My comment was more to Eric or @garrettjonesgoogle as reviewers to think this through.  There are other things that depend on this, someone just needs to check them before approving the changes.  (Internal classes)

jadekler:
We do not support jetty 8 internally. We do support `9.2.10.v20150310`. This PR could possibly be update to that version. That said, I don't have a lot of appetite to take on this without further investigation - as @lesv correctly pointed out, this may affect many dependors.

JustinBeckwith:
Since this is over a year old, should we just close this out for now?

jadekler:
That's a good idea. I think this kind of work would make a lot of sense in some wholesale deprecation of some JDK versions + updating of our dependencies.

PR #157, opened 2017-06-30
moe sync

garrettjonesgoogle:


google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
```diff
@@ -16,18 +16,19 @@
...
-
-import org.mortbay.jetty.Connector;
-import org.mortbay.jetty.Request;
-import org.mortbay.jetty.Server;
-import org.mortbay.jetty.handler.AbstractHandler;
-
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.net.Socket;
+import java.util.concurrent.locks.Condition;
```
ejona86: Why does this add an unused import

garrettjonesgoogle: Cleaned up - this came from a merge mess.

PR #170, opened 2018-02-07
add idToken from token response

emilesalem:
we need to access the id token in order to call google cloud endpoints

i.e
```
GoogleCredential credential = GoogleCredential.getApplicationDefault(); 
credential = credential.createScoped(...); 
credential.refreshToken(); 
credential.getIdToken();
```

### DIFF
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/Credential.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/Credential.java
@@ -129,6 +129,9 @@ public class Credential
   /** Access token issued by the authorization server. */
   private String accessToken;
 
+  /** Id token issued by the authorization server. */
+  private String idToken;
+  
   /**
    * Expected expiration time in milliseconds based on {@link #setExpiresInSeconds} or {@code null}
    * for none.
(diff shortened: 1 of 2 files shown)


emilesalem:
I signed the CLA

JustinBeckwith:
*gentle ping* @google/yoshi-java

chingor13:
Sorry for the delay in getting to this.

I don't think adding `id_token` parsing to `TokenResponse` is the correct approach. `id_token` parsing is already provided by [IdTokenResponse](https://github.com/google/google-oauth-java-client/blob/dev/google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenResponse.java) and [GoogleTokenResponse](https://github.com/google/google-api-java-client/blob/dev/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleTokenResponse.java) from google-api-client. The `TokenResponse` class is a generic implementation for OAuth2 and does not need to include `id_token`.

I'm also unclear on your usage of this with application default credentials as the `id_token` is part of the OpenID Connect protocol on top of 3-legged OAuth 2.0.

For a workaround you can use [GoogleAuthorizationCodeFlow](https://github.com/google/google-api-java-client/blob/dev/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleAuthorizationCodeFlow.java)

PR #173, opened 2018-03-17
Use connector's host in the redirect URI.

mattwhisenhunt:
@lesv PTAL
CC @chanseokoh


lesv:
This seems reasonable.  Someone else is going to take over the maint of this repo when he returns next week.  I'll go over with him all the open PR's.

Issue #142, opened 2017-01-16
Logical issue in computeNonce() method from OAuthParameters class


petermalone:
The computeNonce() suffers from a logical error. As per the method comments, the method "Computes a nonce based on the hex string of a random non-negative long"

The method is as follows:
nonce = Long.toHexString(Math.abs(RANDOM.nextLong()));

RANDOM is defined as a SecureRandom object, and the SecureRandom.nextLong() calls Random.nextLong() which can return positive and negative numbers. 

As per squid:S267 Long.MIN_VALUE could be returned from Random.nextLong(). Calling Math.abs on values returned from these methods is ill-advised.

The problem is that Random.nextLong() can return Long.MIN_VALUE. Due to two's complement representation, the absolute value of Long.MIN_VALUE is itself, the most negative possible long.

My solution is to remove the Math.abs() call and do some shifting, so that we are dealing with positive numbers only - as per the comments of the method.

nonce = Long.toHexString(RANDOM.nextLong() >>> 1);

Apologies for the double commit for this fix.

chingor13:
Sorry for the long delay getting to this.

Would you like to submit your changes as a PR?

petermalone:
Oh wow haha. I forgot completely about this. Yeah let me open a pull request when I'm back at my computer. Should be in less than 2 hours.

petermalone:
I was able to do it quicker than expected. Should be referenced above but here you go: https://github.com/google/google-oauth-java-client/pull/181

If you don't like the format please let me know.

chingor13:
After further consideration, the actual long value and positivity/negativity of the nonce doesn't matter - [it's a random string](https://oauth.net/core/1.0a/#nonce) that is used once. Bitshifting does force the number to be positive but also reduces the number of bits of entropy by 1.

The current implementation can produce `0000000000000000` to `ffffffffffffffff` as the ranges `Long.MIN_VALUE` (`8000000000000000`) to `-1` (`ffffffffffffffff`) and `0` (`0000000000000000`) to `Long.MAX_VALUE` (`7fffffffffffffff`).

If anything, it looks like we can remove the `Math.abs`.

PR #181, opened 2018-07-13
Hotfix/abs.issue

petermalone:
Pull Request for Issue #142 

Apologies for the double commit.

If you'd rather I re-submit as a single commit please let me know.

Contents of issue #142 below:

The computeNonce() suffers from a logical error. As per the method comments, the method "Computes a nonce based on the hex string of a random non-negative long"

The method is as follows:
nonce = Long.toHexString(Math.abs(RANDOM.nextLong()));

RANDOM is defined as a SecureRandom object, and the SecureRandom.nextLong() calls Random.nextLong() which can return positive and negative numbers.

As per squid:S267 Long.MIN_VALUE could be returned from Random.nextLong(). Calling Math.abs on values returned from these methods is ill-advised.

The problem is that Random.nextLong() can return Long.MIN_VALUE. Due to two's complement representation, the absolute value of Long.MIN_VALUE is itself, the most negative possible long.

My solution is to remove the Math.abs() call and do some shifting, so that we are dealing with positive numbers only - as per the comments of the method.

nonce = Long.toHexString(RANDOM.nextLong() >>> 1);

Apologies for the double commit for this fix.

### DIFF
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthParameters.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthParameters.java
@@ -117,7 +117,7 @@ public final class OAuthParameters implements HttpExecuteInterceptor, HttpReques
    * the {@link #nonce} field.
    */
   public void computeNonce() {
-    nonce = Long.toHexString(Math.abs(RANDOM.nextLong()));
+      nonce = Long.toHexString(Math.abs(RANDOM.nextLong() >>> 1));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


chingor13:
After further consideration, the actual long value and positivity/negativity of the nonce doesn't matter - [it's a random string](https://oauth.net/core/1.0a/#nonce) that is used once. Bitshifting does force the number to be positive but also reduces the number of bits of entropy by 1.

The current implementation can produce `0000000000000000` to `ffffffffffffffff` as the ranges `Long.MIN_VALUE` (`8000000000000000`) to `-1` (`ffffffffffffffff`) and `0` (`0000000000000000`) to `Long.MAX_VALUE` (`7fffffffffffffff`).

If anything, it looks like we can remove the `Math.abs`.

Apologies for having you dig up the old code and adding work for  you.

PR #182, opened 2018-07-13
Browser handler

### LINKED ISSUE #175: AuthorizationCodeInstalledApp should not use java.awt.Desktop
AuthorizationCodeInstalledApp uses `java.awt.Deskopt` to [start an external webbrowser](https://github.com/google/google-oauth-java-client/blob/7a861a3e2d2917cf4b9200876ad0716f67bc9c6f/google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java#L129-L135) for user authentication. However, this renders authentication impossible in JavaFX applications as the [application just freezes](https://bugs.openjdk.java.net/browse/JDK-8195086) and no external browser window is opened. In JavaFX, for example, [`WebView`](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/web/WebView.html) might be preferable over an external browser. In general, it is preferable to not have to pull in awt classes when running JavaFX code. 

One way around this would be to pass a handler to `AuthorizationCodeInstalledApp` to allow the caller to decide how to browse a URL.

hanslovsky:
As outlined in #175 it is desirable in some cases to not use the `java.awt.Desktop` to browse the `authorizationUrl` in `AuthorizationCodeInstalledApp`, e.g. when running a `JavaFX` UI. For that purpose, I created a `Browser` interface with a single `browse(String)` method that is passed as an argument to `AuthorizationCodeInstalledApp`. In order not to introduce any breaking changes, there is also a `DefaultBrowser` that simply delegates to `AuthorizationCodeInstalledApp.browse`.

Fixes #175 

Note: `mvn findbugs:check` fails but it also fails on the `dev` branch.

### DIFF
--- a/google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
+++ b/google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
@@ -51,9 +51,9 @@ public class AuthorizationCodeInstalledApp {
     /**
      *
      * @param url url to browse
-     * @throws Exception
+     * @throws IOException
      */
-    public void browse(String url) throws Exception;
+    public void browse(String url) throws IOException;
   }
 
   /**
@@ -64,7 +64,7 @@ public class AuthorizationCodeInstalledApp {
    */
   public static class DefaultBrowser implements Browser{
 
-    public void browse(String url) throws Exception {
+    public void browse(String url) throws IOException {
       AuthorizationCodeInstalledApp.browse(url);
     }
 
@@ -154,13 +154,7 @@ public class AuthorizationCodeInstalledApp {
   protected void onAuthorization(AuthorizationCodeRequestUrl authorizationUrl) throws IOException {
     String url = authorizationUrl.build();
     Preconditions.checkNotNull(url);
-    System.out.println("Please open the following address in your browser:");
-    System.out.println("  " + url);
-    try {
-      browser.browse(url);
-    } catch (Exception e) {
-      LOGGER.log(Level.WARNING, "Unable to open browser", e);
-    }
+    browser.browse(url);
   }
 
   /**
--- a/google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
+++ b/google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
@@ -40,6 +40,30 @@ import java.util.logging.Logger;
  */
 public class AuthorizationCodeInstalledApp {
 
+
+  /**
+   *
+   * Helper interface to allow caller to browse.
+   *
+   */
+  public static interface Browser {
+    public void browse(String url) throws Exception;
+  }
+
+  /**
+   *
+   * Default browser that just throws an exception.
+   *
+   */
+  public static class FailingBrowser implements Browser {
+
+    /** Browse. */
+    public void browse(String url) throws Exception {
+      throw new Exception("Fail");
+    }
+
+  }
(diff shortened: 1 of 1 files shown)


hanslovsky:
I signed it!

google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
```diff
@@ -110,7 +152,15 @@ protected void onAuthorization(AuthorizationCodeRequestUrl authorizationUrl) thr
...
    */
   protected void onAuthorization(AuthorizationCodeRequestUrl authorizationUrl) throws IOException {
-    browse(authorizationUrl.build());
+    String url = authorizationUrl.build();
+    Preconditions.checkNotNull(url);
+    System.out.println("Please open the following address in your browser:");
+    System.out.println("  " + url);
+    try {
+      browser.browse(url);
+    } catch (Exception e) {
```
chingor13: We should be able to remove this try/catch and let an `IOException` thrown by the browser be handled by the caller.

The `DefaultBrowser` already swallows the exception so the current behavior is maintained. A developer may want the browser's exception to be handled by the caller.

google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
```diff
@@ -37,9 +37,39 @@
...
+   * Helper interface to allow caller to browse.
+   *
+   */
+  public static interface Browser {
+    /**
+     *
+     * @param url url to browse
+     * @throws Exception
+     */
+    public void browse(String url) throws Exception;
```
chingor13: Is there a need for this to be `Exception` rather than `IOException`?

google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
```diff
@@ -110,7 +152,15 @@ protected void onAuthorization(AuthorizationCodeRequestUrl authorizationUrl) thr
    * @throws IOException I/O exception
    */
   protected void onAuthorization(AuthorizationCodeRequestUrl authorizationUrl) throws IOException {
-    browse(authorizationUrl.build());
+    String url = authorizationUrl.build();
+    Preconditions.checkNotNull(url);
+    System.out.println("Please open the following address in your browser:");
```
chingor13: We shouldn't need these two `println` statements.

google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
hanslovsky: No particular reason from my side, I will update the interface.

google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
hanslovsky: With the update of the `Browser.browse` signature as suggested in your other comment, this will be possible.

google-oauth-client-java6/src/main/java/com/google/api/client/extensions/java6/auth/oauth2/AuthorizationCodeInstalledApp.java
hanslovsky: Will remove those.

PR #154, opened 2017-05-06
Revert "Master"

lesv:
Reverts google/google-oauth-java-client#153

PR #153, opened 2017-05-04
Master

DebojyotiGit:

PR #186, opened 2018-07-17
Redo documentation fixes for expiresIn vs expirationtime

### LINKED ISSUE #137: Clarify difference between Credential's ExpirationTimeMilliseconds vs ExpiresInSeconds

chingor13:
Fixes #137

This fix was previously contributed in #138 but was merged into master (not the primary branch of this repository). Master was merged into dev (primary branch) without the CLA check passing in #153 and was reverted in #154. Since it was a merge and not a squash and merge, the fixed commits are already in the git history and cannot be cherry-picked in and we cannot revert the revert because of the CLA failure. This this is a new diff with the same contents as #138

### DIFF
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/Credential.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/Credential.java
@@ -389,7 +389,10 @@ public class Credential
     return this;
   }
 
-  /** Expected expiration time in milliseconds or {@code null} for none. */
+  /**
+   * Expected expiration time in milliseconds relative to the
+   * {@link System#currentTimeMillis() Java epoch}, or {@code null} for none.
+   */
   public final Long getExpirationTimeMilliseconds() {
     lock.lock();
     try {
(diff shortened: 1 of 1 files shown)


chingor13:
Create #187

PR #189, opened 2018-07-18
Clarify getUserId's intent in abstract servlets

### LINKED ISSUE #123: getUserId cannot be retrieved without Access Token
AbstractAuthorizationCodeCallbackServlet passes only HttpServletRequest to getUserId(request) method but user info cannot be taken only by it. The problem is that UserInfo could only be retrieved using access token so I suppose it should be passed too. 

```java
String redirectUri = getRedirectUri(req);
TokenResponse response = flow.newTokenRequest(code).setRedirectUri(redirectUri).execute();
String userId = getUserId(req, response.getAccessToken());
Credential credential = flow.createAndStoreCredential(response, userId);
onSuccess(req, resp, credential);
```

For example https://www.googleapis.com/oauth2/v3/userinfo requires access_token parameter.

chingor13:
Fix #123


chingor13:
Whoops, I should have added the donotmerge label. We were waiting until the 1.24.0 release for all PRs.

@jadekler should I revert this and resubmit later?

jadekler:
Let's leave it for now, since in the release process we are no longer thrashing on fresh starts and are beginning to progress beyond the initial clone. If we have to restart the process, we may need to figure something out.

PR #190, opened 2018-07-18
Fix OAuth signature with repeated parameters

### LINKED ISSUE #91: OAuthParameters repeated parameters problem
_From [lautaro....@globant.com](https://code.google.com/u/105025006115827128741/) on April 04, 2014 09:17:00_

It seems like it fails when there is more than one parameter with the same name.

Duplicated parameter names should be taken into account when building the signature according to [9.1.1. Normalize Request Parameters](http://oauth.net/core/1.0/).

But it looks like [OAuthParameters](https://code.google.com/p/google-oauth-java-client/source/browse/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthParameters.java?name=1.17.0-rc) doesn't consider these two possibilities on method "computeSignature()" (line 137). In fact, at line 157 we have the following code:

```
for (Object repeatedValue : (Collection<?>) value) {
    putParameter(parameters, name, repeatedValue);
}
```

which keeps only the last value for a repeated parameter.

Related StackOverflow post: http://stackoverflow.com/questions/4921523/userservice-getcurrentuser-returns-null

_Original issue: http://code.google.com/p/google-oauth-java-client/issues/detail?id=91_

chingor13:
Fixes #91

Uses a SortedMultiset implementation to allow repeated parameter keys. Parameters are sorted first by key, then by value.

### DIFF
--- a/google-oauth-client/pom.xml
+++ b/google-oauth-client/pom.xml
@@ -93,7 +93,6 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <scope>test</scope>
     </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


jadekler:
This is a non-trivial change. Let's make a plan to test this in g3 to see if it'll break things.

chingor13:
This passed internal testing.

Issue #187, opened 2018-07-17
Make master the primary branch


chingor13:
Currently dev is the primary branch, but now that the project is on GitHub, we should make master the primary branch.

jadekler:
Notes:

- This change would need to happen to http and api as well
- There are several places along the way in the release script and possibly `apiserving` that need to be updated as part of this

PR #212, opened 2018-09-18
Be more helpful in the docs for getAccessToken().

ensonic:
Suggest to call refreshToken() if no token was returned.


ensonic:
FYI: It looks two of you travis builder have some configuration issue.

chingor13:
Can you make this PR target the dev branch? For some reason it was setup with dev as the primary branch. We have issue #187 to fix this. CI should run after that.

PR #243, opened 2019-01-07
Cleanup assembly proceses

chingor13:
The generated *.dependencies.html files should not be checked in and should be generated based off of the current list of dependencies.

As part of maven `package` lifecycle, we now generate the dependencies report (via the project-info-reports plugin) and copy them into the assembled zip.


Danielramn1984:
Daniel Ramos Navarro
RFC RAND840816AW0
CIF18010124845

Issue #130, opened 2016-08-16
Upgrade to Jetty 9


elharo:
In particular use org.eclipse.jetty instead of org.mortbay.jetty

chanseokoh:
@ejona86, are you comfortable upgrading to Jetty 9 in `LocalServerReceiver`?

chanseokoh:
Seems like Jetty 9 requires Java 8. My guess is that we can't do this now; I saw Travis build tests for openjdk6 and 7.

evantill:
what about using jetty 8 ? see my PR #155

trejkaz:
We'd like Jetty 9 too. It's just one class, how hard could it be? :)

kermesse2017-gssb:
@trejkaz  I think that the problem is not about complexity but more about risk / side effect (on other projects. And Jetty 9 require Java 8 but  Google-oaut-java-client is compatible with java 6

trejkaz:
What if the Jetty 9 support were done in a java8-only package, leaving alone the current one for people who are still keen on living in the past?

v4dkou:
Encountered this problem when trying to create a service with Spark Framework, which is built around Jetty 9, and as @eevirgen had said, there is a conflict between Servlet 2.* and 3.* APIs

elharo:
Any chance of getting this in now?

jadekler:
Hi @elharo - we're undergoing a release right now as well as cleaning up the backend to make releases in the future easier and quicker. This repo is frozen until that process is done, unfortunately. We'll circle back to this issue and others once the release is done.

cowwoc:
Hi. It's 2019 now and the dependency on this older version of Jetty is turning into a pain. Any update on this issue?

jjYBdx4IL:
This jetty6 dependency is blocking usage of secure versions of jetty. Please, at least factor it out into a separate, optional package so users can choose to drop it or use a newer one. Trying to keep compatibility with out of date java releases is suicide.

Ancient123:
So... There is like a pile of CVE's related to old versions of Jetty at this point. Using a 9 year old version of Jetty with known vulnerabilities seems like a significant security risk... Maybe you guys should bump this up the priority list.

```
[CVE-2011-4461] Cryptographic Issues
[CVE-2017-7656] Improper Access Control
[CVE-2017-7657] Integer Overflow or Wraparound
[CVE-2017-7658] Data Handling
[CVE-2017-9735] Information Exposure
[...]

PR #377, opened 2019-10-04
Use modern-ish jetty

### LINKED ISSUE #233: Upgrade jetty version
See #172.

Ensure that we can upgrade this (internal dependencies).

Ancient123:
Fixes #233 
Works towards #130

### DIFF
--- a/google-oauth-client-jetty/pom.xml
+++ b/google-oauth-client-jetty/pom.xml
@@ -87,8 +87,8 @@
       <artifactId>google-oauth-client-java6</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.mortbay.jetty</groupId>
-      <artifactId>jetty</artifactId>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-server</artifactId>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
--- a/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
+++ b/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
@@ -21,10 +21,10 @@ import java.io.PrintWriter;
 import java.util.concurrent.Semaphore;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import org.mortbay.jetty.Connector;
-import org.mortbay.jetty.Request;
-import org.mortbay.jetty.Server;
-import org.mortbay.jetty.handler.AbstractHandler;
+import org.eclipse.jetty.server.Connector;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.AbstractHandler;
 
 /**
  * OAuth 2.0 verification code receiver that runs a Jetty server on a free port, waiting for a
@@ -118,7 +118,7 @@ public final class LocalServerReceiver implements VerificationCodeReceiver {
     server = new Server(port != -1 ? port : 0);
     Connector connector = server.getConnectors()[0];
     connector.setHost(host);
-    server.addHandler(new CallbackHandler());
+    server.setHandler(new CallbackHandler());
     try {
       server.start();
       port = connector.getLocalPort();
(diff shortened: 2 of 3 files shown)


google-oauth-client-jetty/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>com.google.oauth-client</groupId>
     <artifactId>google-oauth-client-parent</artifactId>
-    <version>1.30.4-SNAPSHOT</version><!-- {x-version-update:google-oauth-client:current} -->
+    <version>1.30.3</version><!-- {x-version-update:google-oauth-client:current} -->
```
elharo: I don't think this should have changed

google-oauth-client-jetty/pom.xml
```diff
@@ -87,8 +87,9 @@
       <artifactId>google-oauth-client-java6</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.mortbay.jetty</groupId>
-      <artifactId>jetty</artifactId>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-server</artifactId>
+      <version>8.2.0.v20160908</version>
```
elharo: maybe a dependencyManagement section somewhere needs to be updated?

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
```diff
@@ -118,7 +118,9 @@ public String getRedirectUri() throws IOException {
     server = new Server(port != -1 ? port : 0);
     Connector connector = server.getConnectors()[0];
     connector.setHost(host);
-    server.addHandler(new CallbackHandler());
+    //Deprecated
```
elharo: delete the commented lines

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
```diff
@@ -254,7 +256,7 @@ public Builder setLandingPages(String successLandingPageUrl, String failureLandi
 
     @Override
     public void handle(
-        String target, HttpServletRequest request, HttpServletResponse response, int dispatch)
+        String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response)
```
elharo: This looks like a change in public API, which we are loath to do. Can the old method be retained and add this one, maybe one with a delegate to the other?

Ancient123: I have no idea how to to that! 
But, I figured after 4 years of there being a request to update a decade old highly vulnerable package ( [CVE-2011-4461][CVE-2017-7656][CVE-2017-7657][CVE-2017-7658][CVE-2017-9735][CVE-2019-10241][CVE-2019-10247] ) I might atleast post a PR to get some traction on the issue.

Ancient123: Also: 
This is public within `class CallbackHandler` which is not public. I am guessing that this is only exposed to the LocalServerReceiver class that it is nested within. Although my knowledge of java is garbage so I may be completely wrong.

google-oauth-client-jetty/pom.xml
Ancient123: Fixed

google-oauth-client-jetty/pom.xml
Ancient123: Found the upstream reference in the main pom.xml and fixed it there.

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
Ancient123: Fixed.

google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
elharo: If the class is non-public, that's OK then.

Ancient123: Looks like there is one style issue I am creating. I will try and fix it. Please re-run kokoro.

Ancient123:
Yay everything passes!

chingor13:
We don't have this version of jetty internally (we do have 9.4.x) so we won't be able to import this in. We should likely update to 9.4 before we attempt to import it internally.

Ancient123:
I am still looking at migrating to 9.4.x but the 8 => 9 upgrade is pretty non-trivial.

PR #438, opened 2020-02-20
deps: update checkstyle plugin

elharo:
@chingor13 @ericraskin This seems necessary to allow us to use try with resources.


chingor13:
Let's turn off checkstyle for CI for Java 7 - we don't need to run it multiple times (the rules are the same).

PR #442, opened 2020-02-24
deps: update animal-sniffer

elharo:
@BenWhitehead


elharo:
We need to turn this off for Java 7.

PR #461, opened 2020-04-15
deps: choose android version of Guava with renovabot

elharo:
@chingor13 I think this is what you suggested yesterday.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,6 +4,12 @@
   ],
   "packageRules": [
     {
+      "packagePatterns": [
+        "^com.google.guava:"
+      ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -3,6 +3,12 @@
     "config:base"
```
chingor13, line 3: IIRC, we needed to pick and choose from things included in config:base to all the versionScheme for guava to take effect: https://github.com/googleapis/java-shared-dependencies/blob/9f844c4b86cfd5e228a1d3b215726d632c21abf6/renovate.json#L2-L12

PR #470, opened 2020-05-29
feat: add PKCE support to AuthorizationCodeFlow

### LINKED ISSUE #469: PKCE support?
I've been looking for information about how to enable PKCE when using the AuthorizationCodeFlow but haven't found anything.

Am I missing something or is PKCE not supported (out of the box) by this library?

StFS:
feat: Add PKCE support to the AuthorizationCodeFlow

According to the OAuth 2.0 for Native Apps RFC, public native clients MUST implement PKCE (https://tools.ietf.org/html/rfc8252#section-6). It is therefor important to add this mechanism to the google-oauth-java-client library. 

- [✓] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-oauth-java-client/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [✓] Ensure the tests and linter pass
- [✓] Code coverage does not decrease (if any source code was changed)
- [✓] Appropriate docs were updated (if necessary)

Fixes #469 ☕️


google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/AuthorizationCodeRequestUrl.java
```diff
@@ -52,6 +54,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
  */
 public class AuthorizationCodeRequestUrl extends AuthorizationRequestUrl {
 
+  /**
+   * The PKCE <a href="https://tools.ietf.org/html/rfc7636#section-4.3">Code Challenge</a>.
+   * @since 1.30.7
```
chingor13, line 59: This will be a minor version bump as we're adding new public surface area to the client, so let's make this 1.31

google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -412,6 +440,71 @@ public final Clock getClock() {
...
+    private String challengeMethod;
+
+    public PKCE() {
+      verifier = generateVerifier();
+      Challenge c = new Challenge(verifier);
+      challenge = c.challenge;
+      challengeMethod = c.method;
+    }
+
+    private String generateVerifier() {
```
chingor13, line 460: nit: this can be static

google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -412,6 +440,71 @@ public final Clock getClock() {
...
+    public String getChallengeMethod() {
+      return challengeMethod;
+    }
+
+    /**
+     * An abstraction to create the PKCE code verifier. It will use the S256 method but
+     * will fall back to using the 'plain' method in the unlikely case that the SHA-256
+     * MessageDigest algorithm implementation can't be loaded.
+     */
+    private static class Challenge {
```
chingor13, line 484: It seems like this functionality could be rolled into the PKCE class as a static method since it's only used during initialization in PKCE

samples/keycloak-pkce-cmdline-sample/src/main/java/com/google/api/services/samples/keycloak/cmdline/PKCESample.java
```diff
@@ -0,0 +1,68 @@
+package com.google.api.services.samples.keycloak.cmdline;
+
+import com.google.api.client.auth.oauth2.*;
```
chingor13, line 3: nit: please import the individual classes

samples/keycloak-pkce-cmdline-sample/src/main/java/com/google/api/services/samples/keycloak/cmdline/PKCESample.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.api.client.extensions.java6.auth.oauth2.AuthorizationCodeInstalledApp;
+import com.google.api.client.extensions.jetty.auth.oauth2.LocalServerReceiver;
+import com.google.api.client.http.*;
```
chingor13, line 6: nit: please import the individual classes

google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -182,8 +194,13 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
    * </pre>
    */
   public AuthorizationCodeRequestUrl newAuthorizationUrl() {
-    return new AuthorizationCodeRequestUrl(authorizationServerEncodedUrl, clientId).setScopes(
-        scopes);
+    AuthorizationCodeRequestUrl acru = new  AuthorizationCodeRequestUrl(authorizationServerEncodedUrl, clientId);
```
chingor13, line 197: nit: Google style is to avoid abbreviations in variable names unless it's a well-known acronym

google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -206,9 +223,20 @@ static TokenResponse requestAccessToken(AuthorizationCodeFlow flow, String code)
    * @param authorizationCode authorization code.
    */
   public AuthorizationCodeTokenRequest newTokenRequest(String authorizationCode) {
+    HttpExecuteInterceptor pkceClientAuthenticationWrapper = new HttpExecuteInterceptor() {
+      @Override
+      public void intercept(HttpRequest request) throws IOException {
+        clientAuthentication.intercept(request);
+        if (pkce != null) {
+          Map<String, Object> data = Data.mapOf(UrlEncodedContent.getContent(request).getData());
+          data.put("code_verifier", pkce.getVerifier());
```
chingor13, line 232: Do we need to re-set the data back on the request? Or is this somehow mutating the data in place?

google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/AuthorizationCodeFlow.java
StFS: I'll create a method from this. It can't easily be static though unless I return a "Pair<String, String>" or a String array of size 2 to represent the challange and method. But a private method that initialises the challenge/method values is plenty good here I think.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/AuthorizationCodeFlow.java
StFS: My understanding is that the request is used only once for the request to obtain the authorization code token.

So I "borrowed" the implementation that is used by the [ClientsParameterAuthentication](https://github.com/googleapis/google-oauth-java-client/blob/6447917c657a5ae4267afbab74dfdb890bbfbf28/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/ClientParametersAuthentication.java#L98) class.

As I said, I believe this object is only used for the one request to get the authorization code and therefore I believe we don't need to reset anything. I'll see if I can perform some experiments to make sure that subsequent requests don't leak this information.

StFS: I experimented a little bit with the `PKCESample` class and after requesting authorization and then creating a request via a transport obtained from the credential object received from the authorization code flow, the code_verifier header is no longer in any requests that are sent out.

So unless I'm misunderstanding your comment, I think we're good.

StFS:
@chingor13 I think I've responded to all your comments and resolved them.

Thank you again so much for taking the time to review this and thanks for your excellent comments. Please let me know if anything else is needed. Should I update the versions.txt file to bump the version or is that just handled by you when you release a new version?

google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -447,23 +447,41 @@ public final Clock getClock() {
...
-    private String generateVerifier() {
+    private static String generateVerifier() {
       SecureRandom sr = new SecureRandom();
       byte[] code = new byte[32];
       sr.nextBytes(code);
       return Base64.encodeBase64URLSafeString(code);
     }
 
+    /**
+     * Create the PKCE code verifier. It will use the S256 method but
```
elharo, line 466: nit: present tense per Google style
i.e. will use the --> uses the
will fall back --> falls back

samples/keycloak-pkce-cmdline-sample/README.md
```diff
@@ -0,0 +1,36 @@
+# Instructions for the Keycloak OAuth2 with PKCE Command-Line Sample
+
+## Browse Online
+
+[Browse Source][browse-source], or main file [DailyMotionSample.java][main-source].
+
+## Command-Line Instructions
+
+**Prerequisites:** install [Java 6 or higher][install-java], [git][install-git], and
```
elharo, line 9: Java 7

samples/keycloak-pkce-cmdline-sample/README.md
```diff
@@ -0,0 +1,36 @@
...
+[Maven][install-maven]. You may need to set your `JAVA_HOME`.
+
+1. Check out the sample code:
```
elharo, line 12: Not sure what keycloak is, or if we can include this part in a google project

samples/keycloak-pkce-cmdline-sample/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+        <version>1.6.8</version>
+        <configuration>
+          <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>8</source>
```
elharo, line 88: 7?

chingor13: Samples are OK to require a higher Java version than the lowest version supported by the library.

samples/keycloak-pkce-cmdline-sample/README.md
StFS: ahh... that had not occurred to me. Keycloak is a JBoss (Red Hat) project, it's an "open source identity and access management solution" (https://www.keycloak.org/) and it's released under the Apache 2.0 license.
So, as such there shouldn't be a problem using it, especially since it's only being used for the sample demonstration and not as an actual component. But maybe there is some other more political/legal reason for it not being possible. I could remove the sample but I would like to keep it as it's pretty good and clear for the purpose of testing and demonstration.

samples/keycloak-pkce-cmdline-sample/pom.xml
StFS: hmm... it seems my IDE snuck this block in, I originally copied the pom from the dailymotion sample so this pom should have been the same as that.

I've removed this compiler-plugin block and it seems to work just fine so I'll commit that.

samples/keycloak-pkce-cmdline-sample/README.md
StFS: Ok, this is copied from the dailymotion sample, that says Java 6 and as per my reply to your comment below, I've removed the config block from the pom that specified a java version explicitly. So should I still change this to 7?

I'll see if I can download Java 6 somewhere and see if everything seems to work if I use that.

elharo: No, the daily motion sample should say 7. Those docs might just be out of date.

samples/keycloak-pkce-cmdline-sample/README.md
elharo: Not a license issue, but we would like the docs for this product to be self-contained.

samples/keycloak-pkce-cmdline-sample/README.md
StFS: ahh... understood, will change the text in the daily motion sample as well.

samples/keycloak-pkce-cmdline-sample/README.md
StFS: gotcha. However, the daily motion example isn't exactly self-contained is is? It actually depends on the user registering an account with dailymotion.com.

But if using keycloak is not acceptable, can you suggest an alternative or should I remove the sample completely?

samples/keycloak-pkce-cmdline-sample/pom.xml
StFS: @chingor13 thanks, that's good to know. I won't go through the trouble of running Java 6 in a linux virtual machine then :)

samples/keycloak-pkce-cmdline-sample/README.md
StFS: @elharo any thoughts on this?

StFS:
@chingor13 what are the next steps here? I believe I've addressed everything you and @elharo have pointed out. The only still outstanding question is whether we can include the PKCE sampe project because it requires the user to run Keycloak in a docker container. I can remove it, but as I said, compared to the other example which requires a user to register with dailymotion.com I would argue that the PKCE sample is more self contained that that.

But mainly I'm eager to get this merged and into a new release so I don't have to create my own custom patch version of the library for us to use ;-)

samples/keycloak-pkce-cmdline-sample/src/main/java/com/google/api/services/samples/keycloak/cmdline/PKCESample.java
```diff
@@ -0,0 +1,74 @@
+package com.google.api.services.samples.keycloak.cmdline;
```
chingor13, line 1: All the source files in the repo need a license header

samples/keycloak-pkce-cmdline-sample/scripts/initialize-keycloak.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
chingor13, line 1: Bash scripts will also need a license header

samples/keycloak-pkce-cmdline-sample/src/main/java/com/google/api/services/samples/keycloak/cmdline/PKCESample.java
StFS: done, along with some more documentation.

samples/keycloak-pkce-cmdline-sample/scripts/initialize-keycloak.sh
StFS: Done, along with more documentation.

StFS:
> Sorry for the delay in reviewing again. Keycloak looks fine to me. The only change we need to make is to ensure the source files have license headers. You can duplicate the other source files for examples.

No worries. Sorry for being so pushy ;)

I've resolved the issue with the license header and added some documentation. Please advise if anything else is needed from me for next steps.

StFS:
@chingor13 just another friendly nudge to get this merged :)

StFS:
Thanks @chingor13 

Just two questions:
1) When would you expect a new version to be released?
2) You commented on the PR that this should be a bump on the minor version since new API is being introduced. Do you keep track of that somewhere (that the next release should be a minor release, not a patch release)?

PR #502, opened 2020-09-18
replace deprecated constructor

elharo:
@lesv


lesv (approved):

lesv:
@elharo Odd - I see it compile, but when I look at the docs https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/net/PercentEscaper.html , I don't see the change in the constructor.

elharo:
This is com.google.api.client.util.PercentEscaper, not Guava's PercentEscaper.

PR #538, opened 2020-10-07
fix: oauth1 signing for url encoded content

### LINKED ISSUE #1: OAuth 1.0: support URL form encoded parameters
_From [yan...@google.com](https://code.google.com/u/111815493751521097215/) on May 11, 2011 10:11:50_

Copied from: https://code.google.com/p/google-api-java-client/issues/detail?id=95 Describe the problem.

Form encoded POST parameters are part of the signature base string and must be signed.  Looking at OAuthAuthorizationHeaderIntercepter it is clear that the interceptor does not do this: it simply passes request method and URL to the OAuthParameters object.  Consequently, the computed signature is wrong and the POST request fails.
This comes up when, e.g. tweeting using the Twitter API:  their API expects a form encoded "status" parameter. GET requests using the OAuthAuthorizationHeaderIntercepter work fine.

How would you expect it to be fixed?

One possible fix would be to provide a second OAuthParameters#computeSignature that accepts a collection of additional parameters that form part of the signature base string.  The OAuthAuthorizationHeaderIntercepter could then check the HttpRequest's method and in the case of a form encoded POST request pass the additional parameters to the #computeSignature method.

Comment 1 by project member yan...@google.com, Jan 8, 2011

Thanks for reporting this issue.  This is actually a known limitation: http://javadoc.google-api-java-client.googlecode.com/hg/1.2.2-alpha/com/google/api/client/auth/oauth/package-summary.html But thanks for bringing it to my attention.  We do want this OAuth library to work properly with the Twitter API.

_Original issue: http://code.google.com/p/google-oauth-java-client/issues/detail?id=1_

cmunden:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-oauth-java-client/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Appropriate docs were updated (if necessary)

Fixes #1 ☕️

### DIFF
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthParameters.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthParameters.java
@@ -280,7 +280,16 @@ public final class OAuthParameters implements HttpExecuteInterceptor, HttpReques
     computeNonce();
     computeTimestamp();
     try {
-      computeSignature(request.getRequestMethod(), request.getUrl());
+      GenericUrl url = request.getUrl();
+      HttpContent content = request.getContent();
+      Map<String, Object> urlEncodedParams = null;
+      if (content instanceof UrlEncodedContent) {
+          urlEncodedParams = Data.mapOf(((UrlEncodedContent) content).getData());
+          url.putAll(urlEncodedParams);
+      }
+      computeSignature(request.getRequestMethod(), url);
+      ofNullable(urlEncodedParams)
+              .ifPresent(contentParams -> contentParams.forEach((key, value) -> url.remove(key)));
     } catch (GeneralSecurityException e) {
       IOException io = new IOException();
       io.initCause(e);
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthParameters.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthParameters.java
@@ -14,16 +14,14 @@
 
 package com.google.api.client.auth.oauth;
 
-import com.google.api.client.http.GenericUrl;
-import com.google.api.client.http.HttpExecuteInterceptor;
-import com.google.api.client.http.HttpRequest;
-import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.client.http.*;
 import com.google.api.client.util.Beta;
+import com.google.api.client.util.Data;
 import com.google.api.client.util.escape.PercentEscaper;
-
 import com.google.common.collect.Multiset;
 import com.google.common.collect.SortedMultiset;
 import com.google.common.collect.TreeMultiset;
+
 import java.io.IOException;
 import java.security.GeneralSecurityException;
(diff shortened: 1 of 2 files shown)


google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthParameters.java
```diff
@@ -14,16 +14,14 @@
 
 package com.google.api.client.auth.oauth;
 
-import com.google.api.client.http.GenericUrl;
-import com.google.api.client.http.HttpExecuteInterceptor;
-import com.google.api.client.http.HttpRequest;
-import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.client.http.*;
```
elharo, line 17: Google doesn't use wildcard imports

elharo (commented):
an issue explaining the bug would help.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthParameters.java
cmunden: I have removed the wildcard import in my most recent commit.  Regarding an issue to explain the bug, a pre-existing issue was referenced in the pull request description.

Fixes #1

elharo:
@silvolu The referenced issue is pretty old. Is this still something we want to do?

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthParametersTest.java
```diff
@@ -72,6 +80,22 @@ public void testSignature() throws GeneralSecurityException {
     assertEquals("GET&https%3A%2F%2Fexample.local&foo%3Dbar%26oauth_signature_method%3Dmock", parameters.signature);
   }
 
+  public void testSignatureWithUrlEncodedContent() throws IOException {
+    OAuthParameters parameters = new OAuthParameters();
+    parameters.signer = new MockSigner();
+
+    GenericUrl url = new GenericUrl("https://example.local?foo=bar");
+    Map<String, Object> contentParameters = Collections.singletonMap("this", (Object) "that");
```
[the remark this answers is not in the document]
cmunden: The cast to Object is necessary in order to create the contentParameters templated to a map of type Map<String, Object>.  Without the explicit cast to Object, the singletonMap method will template to a map of type Map<String, String>.

vladokrsymphony:
Hello, any plans soon to merge this PR?

elharo (approved):

cmunden:
@vladokrsymphony  I recommend that your report your additional issue regarding non ascii characters in the Issues section.  Just go [here](https://github.com/googleapis/google-oauth-java-client/issues/new/choose) and choose "Get Started" in the Bug Report area.

Since this PR is now merged and your bug is of a slightly different nature, then it can be tracked and not missed.  Also remember to provide some good step-by-step instructions on how to recreate the issue.

Happy coding 👍

PR #595, opened 2021-01-11
fix: don't swallow exceptions in LocalServerReceiver

### LINKED ISSUE #503: LocalServerReceiver uses PrintWriter
This causes it to swallow IOExceptions.

elharo:
fixes #503

### DIFF
--- a/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
+++ b/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
@@ -326,6 +326,7 @@ public final class LocalServerReceiver implements VerificationCodeReceiver {
         doc.write("Received verification code. You may now close this window.");
         doc.write("</body>");
         doc.write("</html>\n");
+        doc.flush();
       }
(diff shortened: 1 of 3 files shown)


google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java
```diff
@@ -315,19 +315,19 @@ public void handle(HttpExchange httpExchange) throws IOException {
...
+      try (OutputStream os = exchange.getResponseBody()) {
+        exchange.sendResponseHeaders(HTTP_OK, 0);
+        headers.add("ContentType", "text/html");
+
+        OutputStreamWriter doc = new OutputStreamWriter(os, StandardCharsets.UTF_8);
+        doc.write("<html>");
+        doc.write("<head><title>OAuth 2.0 Authentication Token Received</title></head>");
+        doc.write("<body>");
+        doc.write("Received verification code. You may now close this window.");
+        doc.write("</body>");
```
chingor13, line 327: Not a huge deal as it's writing html, but do we want to preserve the newlines?

chingor13 (approved):

PR #633, opened 2021-02-23
fix: remove Windows CI build

elharo:
This project does not build on Windows due to DataNucleus.


suztomo:
This edit might be overridden by synthtool unless you add this file in the exclusion list:

https://github.com/googleapis/google-oauth-java-client/blob/master/synth.py#L19

elharo (commented):
PTAL

chingor13 (approved):

PR #703, opened 2021-06-18
feat: Added HMAC-SHA256 signature method in OAuthHmacSha256Signer

### LINKED ISSUE #702: OAuthHmacSigner only supports HMAC-SHA1 (no ability to use HMAC-SHA256 which is required by NetSuite)
Description:
As a NetSuite integration developer, I would like to be able to use HMAC-SHA256, so that I can connect to Netsuite when HMAC-SHA1 is deprecated after August 2021.

Solution:
I would like to have the ability to sign OAuth 1.0 requests using HMAC-SHA256.

Other Information:
As the signature method is hard-coded in the OAuthHmacSigner class, the only way to support this would be a new signer class, or a code change to make the signature method configurable.

Additional Information:
NetSuite, a major cloud-based service and ERP, will stop allowing any connections made using HMAC-SHA1 after their next release in August 2021 (exact date TBD).  As such, I would like to have the ability to switch to HMAC-SHA256 as a signature method, so there will be no interruptions in any integrations using this library.

svonduhn:
I have added OAuthHmacSha256Signer to enable using the signature method to HMAC-SHA256.

Unit test/code coverage has been added in OAuthHmacSha256SignerTest.

This change is necessary for establishing connections to NetSuite as they will no longer accept HMAC-SHA1 as a signature method.

Fixes #702  ☕️

This pull request should be close and replaced by #711.  There were issues with the original commit email, and I was unable to resolve them by rebasing (a few of the commits did not get rebased correctly).

### DIFF
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
@@ -24,13 +24,10 @@ import javax.crypto.spec.SecretKeySpec;
 
 /**
  * {@link Beta} <br>
- * OAuth {@code "HMAC-SHA1"} signature method.
- *
- * @since 1.0
- * @author Yaniv Inbar
+ * OAuth {@code "HMAC-SHA256"} signature method.
  */
 @Beta
-public final class OAuthHmacSigner implements OAuthSigner {
+public final class OAuthHmacSha256Signer implements OAuthSigner {
 
   /** Client-shared secret or {@code null} for none. */
   public String clientSharedSecret;
@@ -39,7 +36,7 @@ public final class OAuthHmacSigner implements OAuthSigner {
   public String tokenSharedSecret;
 
   public String getSignatureMethod() {
-    return "HMAC-SHA1";
+    return "HMAC-SHA256";
   }
 
   public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
@@ -56,8 +53,8 @@ public final class OAuthHmacSigner implements OAuthSigner {
     }
     String key = keyBuf.toString();
     // sign
-    SecretKey secretKey = new SecretKeySpec(StringUtils.getBytesUtf8(key), "HmacSHA1");
-    Mac mac = Mac.getInstance("HmacSHA1");
+    SecretKey secretKey = new SecretKeySpec(StringUtils.getBytesUtf8(key), "HmacSHA256");
+    Mac mac = Mac.getInstance("HmacSHA256");
     mac.init(secretKey);
     return Base64.encodeBase64String(mac.doFinal(StringUtils.getBytesUtf8(signatureBaseString)));
   }
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010 Google Inc.
+ * Copyright 2021 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -41,17 +41,17 @@ public final class OAuthHmacSha256Signer implements OAuthSigner {
 
   public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
     // compute key
-    StringBuilder keyBuf = new StringBuilder();
+    StringBuilder keyBuffer = new StringBuilder();
     String clientSharedSecret = this.clientSharedSecret;
     if (clientSharedSecret != null) {
-      keyBuf.append(OAuthParameters.escape(clientSharedSecret));
+      keyBuffer.append(OAuthParameters.escape(clientSharedSecret));
     }
-    keyBuf.append('&');
+    keyBuffer.append('&');
     String tokenSharedSecret = this.tokenSharedSecret;
     if (tokenSharedSecret != null) {
-      keyBuf.append(OAuthParameters.escape(tokenSharedSecret));
+      keyBuffer.append(OAuthParameters.escape(tokenSharedSecret));
     }
-    String key = keyBuf.toString();
+    String key = keyBuffer.toString();
     // sign
     SecretKey secretKey = new SecretKeySpec(StringUtils.getBytesUtf8(key), "HmacSHA256");
     Mac mac = Mac.getInstance("HmacSHA256");
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
@@ -23,41 +23,40 @@ import javax.crypto.SecretKey;
 import javax.crypto.spec.SecretKeySpec;
 
 /**
- * {@link Beta} <br>
  * OAuth {@code "HMAC-SHA256"} signature method.
  */
 @Beta
 public final class OAuthHmacSha256Signer implements OAuthSigner {
 
   /** Client secret */
-  private String clientSharedSecret;
+  private final String clientSharedSecret;
 
   /** Token secret */
   private String tokenSharedSecret;
-
-  public void setClientSecret(String clientSecret) {
-    clientSharedSecret = clientSecret;
-  }
 
   public void setTokenSecret(String tokenSecret) {
     tokenSharedSecret = tokenSecret;
   }
 
+  public OAuthHmacSha256Signer(String clientSecret) {
+    this.clientSharedSecret = clientSecret;
+  }
+
+  @Override
   public String getSignatureMethod() {
     return "HMAC-SHA256";
   }
 
+  @Override
   public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
     // compute key
     StringBuilder keyBuffer = new StringBuilder();
-    String clientSecret = clientSharedSecret;
-    if (clientSecret != null) {
-      keyBuffer.append(OAuthParameters.escape(clientSecret));
+    if (clientSharedSecret != null) {
+      keyBuffer.append(OAuthParameters.escape(clientSharedSecret));
     }
     keyBuffer.append('&');
-    String tokenSecret = tokenSharedSecret;
-    if (tokenSecret != null) {
-      keyBuffer.append(OAuthParameters.escape(tokenSecret));
+    if (tokenSharedSecret != null) {
+      keyBuffer.append(OAuthParameters.escape(tokenSharedSecret));
     }
     String key = keyBuffer.toString();
     // sign
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
@@ -29,11 +29,19 @@ import javax.crypto.spec.SecretKeySpec;
 @Beta
 public final class OAuthHmacSha256Signer implements OAuthSigner {
 
-  /** Client-shared secret or {@code null} for none. */
-  public String clientSharedSecret;
+  /** Client secret */
+  private String clientSharedSecret;
 
-  /** Token-shared secret or {@code null} for none. */
-  public String tokenSharedSecret;
+  /** Token secret */
+  private String tokenSharedSecret;
+
+  public void setClientSecret(String clientSecret) {
+    clientSharedSecret = clientSecret;
+  }
+
+  public void setTokenSecret(String tokenSecret) {
+    tokenSharedSecret = tokenSecret;
+  }
 
   public String getSignatureMethod() {
     return "HMAC-SHA256";
@@ -42,14 +50,14 @@ public final class OAuthHmacSha256Signer implements OAuthSigner {
   public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
     // compute key
     StringBuilder keyBuffer = new StringBuilder();
-    String clientSharedSecret = this.clientSharedSecret;
-    if (clientSharedSecret != null) {
-      keyBuffer.append(OAuthParameters.escape(clientSharedSecret));
+    String clientSecret = clientSharedSecret;
+    if (clientSecret != null) {
+      keyBuffer.append(OAuthParameters.escape(clientSecret));
     }
     keyBuffer.append('&');
-    String tokenSharedSecret = this.tokenSharedSecret;
-    if (tokenSharedSecret != null) {
-      keyBuffer.append(OAuthParameters.escape(tokenSharedSecret));
+    String tokenSecret = tokenSharedSecret;
+    if (tokenSecret != null) {
+      keyBuffer.append(OAuthParameters.escape(tokenSecret));
     }
     String key = keyBuffer.toString();
     // sign
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
@@ -17,7 +17,7 @@ package com.google.api.client.auth.oauth;
 import com.google.api.client.util.Base64;
 import com.google.api.client.util.Beta;
 import com.google.api.client.util.StringUtils;
-import com.google.common.collect.ImmutableMap;
+
 import java.security.GeneralSecurityException;
 import java.util.Map;
 import javax.crypto.Mac;
@@ -55,6 +55,7 @@ public final class OAuthHmacSigner implements OAuthSigner {
     return signatureMethod;
   }
 
+  /** Set the signature method.  Valid signature methods are "HMAC-SHA1" and "HMAC-SHA256" */
   public void setSignatureMethod(String signatureMethod) throws IllegalArgumentException {
     if (signatureMethodMap.containsKey(signatureMethod)) {
       this.signatureMethod = signatureMethod;
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
@@ -17,7 +17,7 @@ package com.google.api.client.auth.oauth;
 import com.google.api.client.util.Base64;
 import com.google.api.client.util.Beta;
 import com.google.api.client.util.StringUtils;
-
+import com.google.common.collect.ImmutableMap;
 import java.security.GeneralSecurityException;
 import java.util.Map;
 import javax.crypto.Mac;
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
@@ -17,7 +17,9 @@ package com.google.api.client.auth.oauth;
 import com.google.api.client.util.Base64;
 import com.google.api.client.util.Beta;
 import com.google.api.client.util.StringUtils;
+import com.google.common.collect.ImmutableMap;
 import java.security.GeneralSecurityException;
+import java.util.Map;
 import javax.crypto.Mac;
 import javax.crypto.SecretKey;
 import javax.crypto.spec.SecretKeySpec;
@@ -32,14 +34,33 @@ import javax.crypto.spec.SecretKeySpec;
 @Beta
 public final class OAuthHmacSigner implements OAuthSigner {
 
+  private final Map<String, String> signatureMethodMap = ImmutableMap.of(
+          "HMAC-SHA1", "HmacSHA1",
+          "HMAC-SHA256", "HmacSHA256"
+  );
+
+  private String signatureMethod = "HMAC-SHA1";
+
   /** Client-shared secret or {@code null} for none. */
   public String clientSharedSecret;
 
   /** Token-shared secret or {@code null} for none. */
   public String tokenSharedSecret;
 
+  private String getAlgorithm() {
+    return signatureMethodMap.get(signatureMethod);
+  }
+
   public String getSignatureMethod() {
-    return "HMAC-SHA1";
+    return signatureMethod;
+  }
+
+  public void setSignatureMethod(String signatureMethod) throws IllegalArgumentException {
+    if (signatureMethodMap.containsKey(signatureMethod)) {
+      this.signatureMethod = signatureMethod;
+    } else {
+      throw new IllegalArgumentException("Signature method " + signatureMethod + " not available");
+    }
   }
 
   public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
@@ -56,8 +77,8 @@ public final class OAuthHmacSigner implements OAuthSigner {
     }
     String key = keyBuf.toString();
     // sign
-    SecretKey secretKey = new SecretKeySpec(StringUtils.getBytesUtf8(key), "HmacSHA1");
-    Mac mac = Mac.getInstance("HmacSHA1");
+    SecretKey secretKey = new SecretKeySpec(StringUtils.getBytesUtf8(key), getAlgorithm());
+    Mac mac = Mac.getInstance(getAlgorithm());
     mac.init(secretKey);
     return Base64.encodeBase64String(mac.doFinal(StringUtils.getBytesUtf8(signatureBaseString)));
   }
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
@@ -26,7 +26,7 @@ import javax.crypto.spec.SecretKeySpec;
 
 /**
  * {@link Beta} <br>
- * OAuth {@code "HMAC-SHA1"} signature method.
+ * OAuth {@code "HMAC-SHA1"} and {@code "HMAC-SHA256"} signature method.
  *
  * @since 1.0
  * @author Yaniv Inbar
--- a/google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
+++ b/google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
@@ -18,18 +18,16 @@ import java.security.GeneralSecurityException;
 import junit.framework.TestCase;
 
 /**
- * Tests {@link OAuthHmacSigner}.
- *
- * @author Yaniv Inbar
+ * Tests {@link OAuthHmacSha256Signer}.
  */
-public class OAuthHmacSignerTest extends TestCase {
+public class OAuthHmacSha256SignerTest extends TestCase {
 
-  private static final String EXPECTED_SIGNATURE = "0anl6O7gtZfslLZ5j3QoTwd0uPY=";
+  private static final String EXPECTED_SIGNATURE = "xDJIQbKJTwGumZFvSG1V3ctym2tz6kD8fKGWPr5ImPU=";
 
   public void testComputeSignature() throws GeneralSecurityException {
-    OAuthHmacSigner signer = new OAuthHmacSigner();
-    signer.clientSharedSecret = "abc";
-    signer.tokenSharedSecret = "def";
-    assertEquals(EXPECTED_SIGNATURE, signer.computeSignature("foo"));
+    OAuthHmacSha256Signer signer = new OAuthHmacSha256Signer();
+    signer.clientSharedSecret = "apiSecret";
+    signer.tokenSharedSecret = "tokenSecret";
+    assertEquals(EXPECTED_SIGNATURE, signer.computeSignature("baseString"));
   }
 }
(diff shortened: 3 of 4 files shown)


google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
```diff
@@ -32,14 +34,34 @@
 @Beta
 public final class OAuthHmacSigner implements OAuthSigner {
 
+  private final Map<String, String> signatureMethodMap = ImmutableMap.of(
+          "HMAC-SHA1", "HmacSHA1",
```
elharo, line 38: I don't understand why you need both the keys and the values here.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
```diff
@@ -32,14 +34,34 @@
...
+  }
+
   public String getSignatureMethod() {
-    return "HMAC-SHA1";
+    return signatureMethod;
+  }
+
+  /** Set the signature method.  Valid signature methods are "HMAC-SHA1" and "HMAC-SHA256" */
+  public void setSignatureMethod(String signatureMethod) throws IllegalArgumentException {
+    if (signatureMethodMap.containsKey(signatureMethod)) {
```
elharo, line 60: Maybe use an enum instead?

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSignerTest.java
```diff
@@ -32,4 +35,39 @@ public void testComputeSignature() throws GeneralSecurityException {
     signer.tokenSharedSecret = "def";
     assertEquals(EXPECTED_SIGNATURE, signer.computeSignature("foo"));
   }
+
+  public void testComputeSignatureHmacSha256() throws GeneralSecurityException {
+    final OAuthHmacSigner signer = new OAuthHmacSigner();
```
elharo, line 40: Google style specifies no final on local variables except where required

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSignerTest.java
```diff
@@ -32,4 +35,39 @@ public void testComputeSignature() throws GeneralSecurityException {
...
+  public void testGetSignatureMethodHmacSha256() {
+    final OAuthHmacSigner signer = new OAuthHmacSigner();
+    final String signatureMethod = "HMAC-SHA256";
+    signer.setSignatureMethod(signatureMethod);
+    assertEquals(signatureMethod, signer.getSignatureMethod());
+  }
+
+  public void testSetSignatureMethodHmacMD5() {
+    assertThrows(
+            "Signature method HMAC-MD5 not available",
```
elharo, line 63: four space indents for continuation lines, per Google style guide

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
```diff
@@ -32,14 +34,34 @@
 @Beta
 public final class OAuthHmacSigner implements OAuthSigner {
```
elharo, line 35: class doc comment needs to be updated

elharo (commented):
perhaps HMAC-SHA256 should be a separate class? I'm not sure.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
svonduhn: Good question.  The signature method that need to be used for the OAuth 1.0 header has to be of the format:  "HMAC-SHA1" or "HMAC-SHA2", while the actual Algorithm strings that Java requires have to be of the format "HmacSHA1" or "HmacSHA256" as described here:  https://docs.oracle.com/javase/9/docs/specs/security/standard-names.html#mac-algorithms.

My intent was to use a Map to associate the signature method as the key with the corresponding algorithm as the value, so there would not be an accidental mismatch.

svonduhn:
> perhaps HMAC-SHA256 should be a separate class? I'm not sure.

I think you may be right about having a separate class.  I was trying not to repeat code in a class that would be similar in every way except for three lines, but having a separate class makes a lot of sense, particularly since this is likely to be a special use case signing class, only needed for a few services.  I am going to refactor this as a separate class and see if that is simpler.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
svonduhn: this has been reverted back in favor of a new class

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
svonduhn: this has been reverted back in favor of a new class

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSignerTest.java
svonduhn: this has been reverted back in favor of a new class

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (c) 2010 Google Inc.
```
elharo, line 2: Needs current copyright date, Inc. --> LLC 

and never, ever use (c) in anything, here or elsewhere. It is not recognized by the law, no matter how many places you find it.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,61 @@
...
+import java.security.GeneralSecurityException;
+import javax.crypto.Mac;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+
+/**
+ * {@link Beta} <br>
+ * OAuth {@code "HMAC-SHA256"} signature method.
+ */
+@Beta
```
elharo, line 29: auth leader (not me) needs to figure out if we want this and if it should be beta

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,61 @@
...
+public final class OAuthHmacSha256Signer implements OAuthSigner {
+
+  /** Client-shared secret or {@code null} for none. */
+  public String clientSharedSecret;
```
elharo, line 33: private on fields please

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /** Token-shared secret or {@code null} for none. */
+  public String tokenSharedSecret;
```
elharo, line 36: ditto, also make final

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public String getSignatureMethod() {
+    return "HMAC-SHA256";
+  }
+
+  public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
+    // compute key
+    StringBuilder keyBuf = new StringBuilder();
```
elharo, line 44: keyBuffer

Google style never abbreviates

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) 2012 Google Inc.
```
elharo, line 2: fix copyright

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.api.client.auth.oauth;
+
+import java.security.GeneralSecurityException;
+import junit.framework.TestCase;
```
elharo, line 18: org.junit preferred

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: I can make these fields private, but then will also need to add accessors/mutators.  I noticed in the original class (OAuthHmacSigner) these fields were public and had methods to get/set the values.  Do we want to have this signing class implemented differently from the other on that it is based on?  Existing code using OAuthHmacSigner that needs to migrate to this class would have fewer code changes keeping it the same, but then the API was labelled Beta.

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
svonduhn: The POM file only has the current junit dependency (junit:junit:4.13.2).  I can switch to a different junit, but that would require adding a new dependency.  Which group/artifact/version org.junit package is the preferred one to add and use?

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: this has been updated

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: this has been updated

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
svonduhn: this has been updated

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
elharo: mutable public fields are a hard no. Don't try to be consistent with bad code. You can probably make these final with no mutators.

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
elharo: junit.* is the old legacy classes. org.junit are the new ones from JUnit 4.

elharo (changes requested):

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: The client secret and the token secret probably shouldn't be final, as there is often a multi-step flow to obtain the values in separate steps, so they need to be set after the signer instance is created in most instances.

Also, in looking at the other code required for making a call, the `OAuthParameters` class will still have mutable public fields (similar to `OAuthHmacSigner`), and it is necessary to use that class together with the signer, to make a complete OAuth 1.0 signed request:

Current code needed for HMAC-SHA1:
```
OAuthHmacSigner hmacSha1Signer = new OAuthHmacSigner();
hmacSha1Signer.clientSharedSecret = CLIENT_SECRET;
hmacSha1Signer.tokenSharedSecret = TOKEN_SECRET;
OAuthParameters oauthParameters = new OAuthParameters();
oauthParameters.signer = hmacSha1Signer;
oauthParameters.consumerKey = CONSUMER_KEY;
oauthParameters.token = ACCESS_TOKEN;
HttpRequestFactory requestFactory = new NetHttpTransport().createRequestFactory(oauthParameters);
```

Making the fields on the new class private with setters would make the code look like this:

New code needed for HMAC-SHA256:
```
OAuthHmacSha256Signer hmacSha256Signer = new OAuthHmacSha256Signer();
hmacSha256Signer.setClientSecret(CLIENT_SECRET);
hmacSha256Signer.setTokenSecret(TOKEN_SECRET);
OAuthParameters oauthParameters = new OAuthParameters();
oauthParameters.signer = hmacSha256Signer;
oauthParameters.consumerKey = CONSUMER_KEY;
oauthParameters.token = ACCESS_TOKEN;
HttpRequestFactory requestFactory = new NetHttpTransport().createRequestFactory(oauthParameters);
```

It feels a little bit inconsistent this way, even though this is arguably the better way.  Should anything be done about `OAuthParameters` ?  I'd normally not change something if it isn't necessary.

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
svonduhn: Thank you, I misunderstood.  I updated the test to use org.junit / JUnit4

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: I made the fields private, with setter methods.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: I made the field private, but not final, due to OAuth 1.0 frequently setting these values in multiple steps of the flow, so they cannot always be set during construction.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: Since the interface the `OAuthSigner` interface itself is Beta I took that as a signal that any classes implementing it should also be Beta.

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSignerTest.java
svonduhn: this has been reverted back in favor of a new class

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSigner.java
svonduhn: this has been reverted back in favor of a new class

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
elharo: Either way annotation might be enough. Is there a reason to include this in doc comment too?

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public void setClientSecret(String clientSecret) {
+    clientSharedSecret = clientSecret;
+  }
+
+  public void setTokenSecret(String tokenSecret) {
+    tokenSharedSecret = tokenSecret;
+  }
+
+  public String getSignatureMethod() {
```
elharo, line 46: `@Override`

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,69 @@
...
+    return "HMAC-SHA256";
+  }
+
+  public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
```
elharo, line 50: `@Override`

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,69 @@
...
+@Beta
+public final class OAuthHmacSha256Signer implements OAuthSigner {
+
+  /** Client secret */
+  private String clientSharedSecret;
+
+  /** Token secret */
+  private String tokenSharedSecret;
+
+  public void setClientSecret(String clientSecret) {
```
elharo, line 38: Why are these mutable? It makes the class not thread-safe.

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Tests {@link OAuthHmacSha256Signer}.
+ */
+public class OAuthHmacSha256SignerTest {
+
+  private static final String EXPECTED_SIGNATURE = "xDJIQbKJTwGumZFvSG1V3ctym2tz6kD8fKGWPr5ImPU=";
+
+  @Test
+  public void testComputeSignature() throws GeneralSecurityException {
```
elharo, line 30: if mutability is deliberate it should be tested.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,69 @@
...
+  public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
+    // compute key
+    StringBuilder keyBuffer = new StringBuilder();
+    String clientSecret = clientSharedSecret;
+    if (clientSecret != null) {
+      keyBuffer.append(OAuthParameters.escape(clientSecret));
+    }
+    keyBuffer.append('&');
+    String tokenSecret = tokenSharedSecret;
+    if (tokenSecret != null) {
```
elharo, line 59: this branch is not tested when tokenSecret is null

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public String getSignatureMethod() {
+    return "HMAC-SHA256";
+  }
+
+  public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
+    // compute key
+    StringBuilder keyBuffer = new StringBuilder();
+    String clientSecret = clientSharedSecret;
+    if (clientSecret != null) {
```
elharo, line 54: this branch is not tested when clientSecret is null

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  public String getSignatureMethod() {
+    return "HMAC-SHA256";
+  }
+
+  public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
+    // compute key
+    StringBuilder keyBuffer = new StringBuilder();
+    String clientSecret = clientSharedSecret;
```
elharo, line 53: I don't see why you copy into a local variable here.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,69 @@
...
+    if (clientSecret != null) {
+      keyBuffer.append(OAuthParameters.escape(clientSecret));
+    }
+    keyBuffer.append('&');
+    String tokenSecret = tokenSharedSecret;
```
elharo, line 58: I don't see why you copy into a local variable here.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: The client secret can be final, so I did change it, however the token secret usually is not known until after the first request of the three-legged OAuth 1.0 authorization flow.  Because of that, it needs to be able to be set later.  This is common for OAuth 1.0, as it is a multi-step process where not all of the values are known prior to the first step.

The token secret will usually be set to the final value only once during the authorization flow, before the second step, and never changed again.

google-oauth-client/src/test/java/com/google/api/client/auth/oauth/OAuthHmacSha256SignerTest.java
svonduhn: New tests have been added to cover the null cases

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: New tests have been added to cover the cases when tokenSecret is null

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: New tests have been added to cover the cases when clientSecret is null

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: No need, it was simply in the original class that way.  I have eliminated the local variable.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: No need, it was simply in the original class that way.  I have eliminated the local variable.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: This has been updated.  Strangely, the annotation did not exist in the original implementing class.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: This has been updated.  Strangely, the annotation did not exist in the original implementing class.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,68 @@
...
+public final class OAuthHmacSha256Signer implements OAuthSigner {
+
+  /** Client secret */
+  private final String clientSharedSecret;
+
+  /** Token secret */
+  private String tokenSharedSecret;
+
+  public void setTokenSecret(String tokenSecret) {
+    tokenSharedSecret = tokenSecret;
```
elharo, line 38: This should probably fail if the existing field is not null. Or does it need to be reset? What do you think?

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public OAuthHmacSha256Signer(String clientSecret) {
+    this.clientSharedSecret = clientSecret;
```
elharo, line 42: does it make sense for this field to be null?

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public String computeSignature(String signatureBaseString) throws GeneralSecurityException {
+    // compute key
+    StringBuilder keyBuffer = new StringBuilder();
+    if (clientSharedSecret != null) {
+      keyBuffer.append(OAuthParameters.escape(clientSharedSecret));
+    }
+    keyBuffer.append('&');
+    if (tokenSharedSecret != null) {
```
elharo, line 58: It's OK for this to be omitted?

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: removed this from the doc comment

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: There is a legitimate case for swapping a temporary token for a permanent token.  That is a common scenario in OAuth 1.0

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: In reality, no, this should never be null.  I have worked with other OAuth libraries, though, and none of them actually validate that this is non-null.  However, if any code sets this to be null, then the OAuth calls will be completely non-functional, failing to authenticate.  Setting this to a valid string is critical to OAuth working at all.  None of the other signer classes do any validation of these (in fact, they are just public variables), and we could be more restrictive here, but in practice this will have to be set to a vlid value or the authentication simply won't work.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
svonduhn: Yes, the token secret can be null, a blank string, or even a temporary token for the first call of the authorization flow.  It is intended to be replaced/updated with a permanent token after the initial call.

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
elharo: OK. Can you add a test for this?

TimurSadykov:
@svonduhn Please sign the sla, it is required for all the contributors. Here is the instruction: https://opensource.google/docs/cla/#external-contributors

svonduhn:
> @svonduhn Please sign the cla, it is required for all the contributors. Here is the instruction: https://opensource.google/docs/cla/#external-contributors

@TimurSadykov I signed the CLA here:  https://cla.developers.google.com/clas/edit?id=111692684224848719787&kind=KIND_INDIVIDUAL&domain=DOMAIN_GOOGLE

I also added the Google CLA GitHub App to my GitHub account, but I am not sure what else needs to be done to get the pull request to recognize the CLA.  Is there something additional needed?

svonduhn:
@googlebot rescan

TimurSadykov:
@svonduhn Hey, sorry for delay. I see what is happening. I can see that your GitHub login has CLA signed. However, it seems like you made a commit using different email, that is not associated with your account: svonduhn@commercehub.com

what you can try to do is to add that email to you GitHub account and then rescan?

svonduhn:
@TimurSadykov you are right, the commits were made with the wrong email address.  The only problem I have now is that when I try to add that email address to my account, it says that it is already associated with a different Google account (which it should not be, but that is the error I am getting).  Because of that, it seems that I can't actually fix the issue.

Would it be reasonable (acceptable) to simply create a new pull request with these changes using the correct email address, and close this one?  My apologies for the trouble.  I have updated my .gitconfig to prevent this from happening in the future.

Let me know how best to proceed.  Thank you.

svonduhn:
@elharo @TimurSadykov This pull request should be close and replaced by #711.  There were issues with my original commit email, and I was unable to resolve them via an interactive rebase (a few of the commits did not get rebased correctly).

I do apologize for the trouble.  The new pull request does appear to be passing the CLA check now.  Thank you for your time!

svonduhn:
closing, as this was included in PR #711, which has been merged.

PR #711, opened 2021-08-05
feat: Add HMAC-SHA256 signature method for OAuth 1.0

### LINKED ISSUE #702: OAuthHmacSigner only supports HMAC-SHA1 (no ability to use HMAC-SHA256 which is required by NetSuite)

svonduhn:
I have added OAuthHmacSha256Signer to enable using the signature method HMAC-SHA256.

Unit test/code coverage has been added in OAuthHmacSha256SignerTest.

This change is necessary for establishing connections to NetSuite as they will no longer accept HMAC-SHA1 as a signature method.

Fixes #702

This replaces pull request #703, which should be closed in favor of this one.  The code is the same (two of the four files modified in the original PR were reverted back to their original versions), so there are only two new files, one class and a test class.

### DIFF
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
@@ -16,15 +16,12 @@ package com.google.api.client.auth.oauth;
 
 import com.google.api.client.util.StringUtils;
 import com.google.common.io.BaseEncoding;
-
 import java.security.GeneralSecurityException;
 import javax.crypto.Mac;
 import javax.crypto.SecretKey;
 import javax.crypto.spec.SecretKeySpec;
 
-/**
- * OAuth {@code "HMAC-SHA256"} signature method.
- */
+/** OAuth {@code "HMAC-SHA256"} signature method. */
 public final class OAuthHmacSha256Signer implements OAuthSigner {
 
   /** Client secret */
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
@@ -14,9 +14,9 @@
 
 package com.google.api.client.auth.oauth;
 
-import com.google.api.client.util.Base64;
-import com.google.api.client.util.Beta;
 import com.google.api.client.util.StringUtils;
+import com.google.common.io.BaseEncoding;
+
 import java.security.GeneralSecurityException;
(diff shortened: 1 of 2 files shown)


TimurSadykov (approved):

google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthHmacSha256Signer.java
```diff
@@ -0,0 +1,68 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.api.client.auth.oauth;
+
+import com.google.api.client.util.Base64;
+import com.google.api.client.util.Beta;
```
TimurSadykov, line 18: those are deprecated (internally uses com.google.common.io.BaseEncoding), switch to com.google.common.io.BaseEncoding directly to cleanup warnings?

svonduhn: This change has been made

TimurSadykov:
@svonduhn you can run "mvn com.coveo:fmt-maven-plugin:format" to fix linter issues

svonduhn:
@TimurSadykov thank you for that command.  Those issues should now be fixed as well.

TimurSadykov:
@svonduhn There is still a linter error, but seems like unrelated to your change, i'm looking into it..

Neenu1995 (approved):

TimurSadykov:
@svonduhn linter is fixed, you can merge if you are ready
We will release shortly after

svonduhn:
@TimurSadykov Everything seems fine to me.  I don't have access to merge the request, but it looks like it is ready for whoever is able to merge it.

TimurSadykov:
@svonduhn thanks for contributing!

PR #733, opened 2021-08-23
chore: removing the assembly module from the BOM

### LINKED ISSUE #732: Shall we drop "google-oauth-client-assembly" artifact from the BOM?
https://github.com/googleapis/java-shared-dependencies/pull/446 fails because google-oauth-client-assembly artifact is not declared with "pom" packaging in the BOM. I need either drop the entry from the BOM or declare it as "pom" packaging appropriately.

suztomo:
@chingor13 Do you know the purpose of this assembly module in the BOM? https://repo1.maven.org/maven2/com/google/oauth-client/google-oauth-client-assembly/1.31.5/ shows there's a zip file uploaded to Maven Central. I believe this is not to be used via the BOM. If you're ok, then let's delete the entry from the BOM. The entry is blocking me to apply Linkage Checker enforcer rule to the shared dependencies BOM (https://github.com/googleapis/java-shared-dependencies/pull/446#issuecomment-904098003)

---

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-oauth-java-client/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Appropriate docs were updated (if necessary)

Fixes #732 ☕️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,18 +1,19 @@
 # Changelog
 
-### [1.32.2](https://www.github.com/googleapis/google-oauth-java-client/compare/v1.32.1...v1.32.2) (2021-08-23)
-
-
-### Bug Fixes
-
-* fix release ([#728](https://www.github.com/googleapis/google-oauth-java-client/issues/728)) ([855503a](https://www.github.com/googleapis/google-oauth-java-client/commit/855503afa55def3192e2bdaf7247aa2484a2abbd))
-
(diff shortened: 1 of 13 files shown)


suztomo:
@timursadykov @silvolu (I got your names from Neenu) Do you know anything about [this google-oauth-client-assembly](https://repo1.maven.org/maven2/com/google/oauth-client/google-oauth-client-assembly/1.32.1/) artfiact?

It seems that purpose of the module is to generate the zip file (of which I don't know the purpose). If so, google-oauth-client-assembly shouldn't have been added to the google-oauth-client-bom/pom.xml. What do you think?

PR #816, opened 2022-01-19
chore: downstream check for all libraries in single job

Neenu1995:
Test core java libraries in java client libraries before release. Feature/code changes introduced in each PR can be tested on downstream java client libraries by adding `downstream-check:run` label. For now, unit tests are being run.

Next phase involves running samples and integration tests.

Note: Some of these tests are expected to fail due to individual client library conditions and do block the merging of this PR.

### DIFF
--- a/.kokoro/downstream-client-library-check.sh
+++ b/.kokoro/downstream-client-library-check.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


.kokoro/downstream-client-library-check.sh
```diff
@@ -0,0 +1,104 @@
...
+
+
+CORE_LIBRARY_ARTIFACT=$1
+CLIENT_LIBRARY=$2
+## Get the directory of the build script
+scriptDir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+## cd to the parent directory, i.e. the root of the git repo
+cd ${scriptDir}/..
+
+# Make java core library artifacts available for 'mvn validate' at the bottom
```
lesv, line 28: `mvn verify`

lesv (approved):
nit

Also mention about individual failures

lesv (approved):

PR #861, opened 2022-03-25
fix: add signature verification to IdTokenVerifier

### LINKED ISSUE #786: IdTokenVerifier does not verify the signature of ID Token
https://github.com/googleapis/google-oauth-java-client/blob/5c202d654a33f6113de6151d383b8d0f88fab14b/google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java#L152

The _verify_ method in _IdTokenVerifier_ does not validate the signature before verifying the claims (e.g., _iss_, _aud_, etc.). This is also a violation of the ID Token Validation steps mentioned in the current OpenID Connect Spec (https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation). 

The spec requires to validate the _signature_ of ID token for apps that cannot guarantee TLS communication, which is the case for this library. This library initiates a local server that can run on any client machine without TLS support. So, it is critical to validate the signature, before trusting the claims of an ID token, which can be received from a malicious service provider. 

Thanks.

TimurSadykov:
Porting the functionality from  google-auth-library-java, see https://github.com/googleapis/google-auth-library-java/blob/1434839e704621dc21f902c81cb67e66f7ef1a05/oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java#L81

Fixes #786 ☕️


google-oauth-client/src/test/java/com/google/api/client/auth/openidconnect/IdTokenVerifierTest.java
```diff
@@ -151,4 +184,166 @@ public void testMissingAudience() {
...
+            .setHttpTransportFactory(httpTransportFactory)
+            .build();
+
+    try {
+      tokenVerifier.verify(IdToken.parse(JSON_FACTORY, ES256_TOKEN));
+    } catch (VerificationException ex) {
+      assertTrue(ex.getMessage().contains("Error fetching PublicKey"));
+      return;
+    }
+    throw new Exception("Should have failed verification");
```
suztomo, line 224: My 2 cent:

I would put `fail("Should have failed verification");` immediately after tokenVerifier.verify

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -141,10 +194,60 @@ public final String getIssuer() {
...
+    if (certificatesLocation != null) return certificatesLocation;
+
+    switch (header.getAlgorithm()) {
+      case "RS256":
+        return FEDERATED_SIGNON_CERT_URL;
+      case "ES256":
+        return IAP_CERT_URL;
+    }
+
+    throw new VerificationException("Unknown algorithm");
```
suztomo, line 250: Can you add the value of the algorithm in the exception? It would help troubleshooting.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -141,10 +194,60 @@ public final String getIssuer() {
...
+    }
+
+    if (Boolean.parseBoolean(environment.getVariable(SKIP_SIGNATURE_ENV_VAR))) {
+      return true;
+    }
+
+    // Short-circuit signature types
+    if (!SUPPORTED_ALGORITHMS.contains(idToken.getHeader().getAlgorithm())) {
+      throw new VerificationException(
+          "Unexpected signing algorithm: expected either RS256 or ES256");
```
suztomo, line 214:
```suggestion
          "Unexpected signing algorithm: expected either RS256 or ES256 but got " + idToken.getHeader().getAlgorithm());
```

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +402,163 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
+    }
+
+    /**
+     * Set the HttpTransportFactory used for requesting public keys from the certificate URL. Used
+     * mostly for testing.
+     *
+     * @param httpTransportFactory the HttpTransportFactory used to build certificate URL requests
+     * @return the builder
+     */
+    public Builder setHttpTransportFactory(HttpTransportFactory httpTransportFactory) {
```
suztomo, line 424: > Used mostly for testing.

Would you check whether this needs to be public?

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +402,163 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
       return this;
     }
+
+    /** Returns an instance of the {@link Environment} */
+    public final Environment getEnvironment() {
+      return environment;
+    }
+
+    /** Sets the environment. Used mostly for testing */
+    public Builder setEnvironment(Environment environment) {
```
suztomo, line 412: Would you check whether this needs to be public?

suztomo:
Lint is failing:

```
Error:  Found 7 non-complying files, failing build
Error:  To fix formatting errors, run "mvn com.coveo:fmt-maven-plugin:format"
Error:  Non complying file: /home/runner/work/google-oauth-java-client/google-oauth-java-client/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/AuthorizationCodeResponseUrl.java
Error:  Non complying file: /home/runner/work/google-oauth-java-client/google-oauth-java-client/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/StoredCredential.java
Error:  Non complying file: /home/runner/work/google-oauth-java-client/google-oauth-java-client/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthGetAccessToken.java
Error:  Non complying file: /home/runner/work/google-oauth-java-client/google-oauth-java-client/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/AbstractOAuthGetToken.java
Error:  Non complying file: /home/runner/work/google-oauth-java-client/google-oauth-java-client/google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
Error:  Non complying file: /home/runner/work/google-oauth-java-client/google-oauth-java-client/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthAuthorizeTemporaryTokenUrl.java
Error:  Non complying file: /home/runner/work/google-oauth-java-client/google-oauth-java-client/google-oauth-client/src/main/java/com/google/api/client/auth/oauth/OAuthGetTemporaryToken.java
```

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +402,163 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
+      return CertificateFactory.getInstance("X.509")
+          .generateCertificate(new ByteArrayInputStream(publicPem.getBytes("UTF-8")))
+          .getPublicKey();
+    }
+
+    private PublicKey buildRs256PublicKey(JsonWebKey key)
+        throws NoSuchAlgorithmException, InvalidKeySpecException {
+      com.google.common.base.Preconditions.checkArgument("RSA".equals(key.kty));
+      com.google.common.base.Preconditions.checkNotNull(key.e);
+      com.google.common.base.Preconditions.checkNotNull(key.n);
```
suztomo, line 527: Memo for myself `com.google.api.client.util.Preconditions` is conflicting.

(But maybe static import would work?)

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -141,10 +194,60 @@ public final String getIssuer() {
    * @param idToken ID token
    * @return {@code true} if verified successfully or {@code false} if failed
    */
-  public boolean verify(IdToken idToken) {
-    return (issuers == null || idToken.verifyIssuer(issuers))
-        && (audience == null || idToken.verifyAudience(audience))
-        && idToken.verifyTime(clock.currentTimeMillis(), acceptableTimeSkewSeconds);
+  public boolean verify(IdToken idToken) throws VerificationException {
```
suztomo, line 197: Would you improve the Javadoc that explains the newly-added case?

google-oauth-client/src/test/java/com/google/api/client/auth/openidconnect/IdTokenVerifierTest.java
```diff
@@ -39,6 +56,25 @@
   private static final String ISSUER2 = ISSUER + "2";
   private static final String ISSUER3 = ISSUER + "3";
 
+  private static final String ES256_TOKEN =
```
suztomo, line 59: Would you add source code comment about how to generate the values?

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +402,163 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
+      try {
+        HttpRequest request =
+            httpTransport
+                .createRequestFactory()
+                .buildGetRequest(new GenericUrl(certificateUrl))
+                .setParser(GsonFactory.getDefaultInstance().createJsonObjectParser());
+        HttpResponse response = request.execute();
+        jwks = response.parseAs(JsonWebKeySet.class);
+      } catch (IOException io) {
+        return ImmutableMap.of();
```
suztomo, line 480: This doesn't show any error or exceptions when the certificate doesn't have the expected format. Is there.a good way to report the problem?

google-oauth-client/src/test/java/com/google/api/client/auth/openidconnect/IdTokenVerifierTest.java
TimurSadykov: Those are some test examples, not generated ones as I understand.

google-oauth-client/pom.xml
```diff
@@ -93,5 +93,9 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.http-client</groupId>
+      <artifactId>google-http-client-gson</artifactId>
+    </dependency>
```
chingor13, line 99: There's already an entry for google-http-client-gson on line 82-86 with test scope - can drop the test scope

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/Environment.java
```diff
@@ -0,0 +1,7 @@
+package com.google.api.client.auth.openidconnect;
```
chingor13, line 1: This file needs the license header.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
chingor13: These ones should be package private - `Environment` interface is already package private.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -76,11 +123,17 @@ public IdTokenVerifier() {
...
+    this.certificatesLocation = builder.certificatesLocation;
     clock = builder.clock;
     acceptableTimeSkewSeconds = builder.acceptableTimeSkewSeconds;
     issuers = builder.issuers == null ? null : Collections.unmodifiableCollection(builder.issuers);
     audience =
         builder.audience == null ? null : Collections.unmodifiableCollection(builder.audience);
+    this.publicKeyCache =
+        CacheBuilder.newBuilder()
+            .expireAfterWrite(1, TimeUnit.HOURS)
+            .build(new PublicKeyLoader(builder.httpTransportFactory));
```
chingor13, line 135: `httpTransportFactory` is now required or you'll get a NPE later. We generally prefer to use preconditions to throw the NPE early for required attributes (using guava's `Preconditions.checkNotNull`)

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
chingor13: We don't appear to have a default http transport, so the user needs to set one. We could potentially set a default (using java.net transport). This is also important for picking which http transport library to use (and configuring proxies and such)

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
chingor13: I'm surprised the breaking change detector doesn't see adding a checked exception as a breaking change.

It also feels a bit weird to have 2 different mechanisms for indicating an invalid token. In the google-auth-library implementation, we only throw `VerificationException`s (to be able to communicate why it failed). This legacy implementation already is hiding why the token in invalid.

chingor13: Another option, in line with Tomo's suggestion is to add a new `verifySignature` method that throws `VerificationException`. `verify` could call `verifySignature` and catch the exception and return `false` (to not change the interface)

chingor13 (changes requested):

google-oauth-client/pom.xml
TimurSadykov: oh missed that, thanks!

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: This definitely makes sense, I left it as is since it looks as intentional. I tried to figure out the background for this and it looks like - it is related to LoadingCache implementation, like unable to throw checked exception from runner or something like that. My call is to leave this specific one to avoid as is because: 
1. it does not seem to be a pain point for other library
2. we want to error on the side of reliability at the moment, we can improve experience later
3. library is already officially deprecated

@chingor13 are you ok with that?

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
suztomo: Good catch. If I understand correctly, without a default, the existing code all fails with NPE. That's not good. I want this change to work for most (expected) use cases without any code modification.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/Environment.java
TimurSadykov: yep, fixed

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: Good catch, thanks, I think here we want to set a default ourselves in case of null, otherwise it is going to be a breaking change?

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: fixed, thx!

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: I will set to nettransport by default. And I will add extended error message in case cert failed to fetch pointing to transport and a variable to disable

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: Yes, its a test infra, thanks!

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -141,10 +194,60 @@ public final String getIssuer() {
...
+      throw new VerificationException(
+          "Unexpected signing algorithm: expected either RS256 or ES256");
+    }
+
+    PublicKey publicKeyToUse = null;
+    try {
+      String certificateLocation = getCertificateLocation(idToken.getHeader());
+      publicKeyToUse = publicKeyCache.get(certificateLocation).get(idToken.getHeader().getKeyId());
+    } catch (ExecutionException | UncheckedExecutionException e) {
+      throw new VerificationException("Error fetching PublicKey from certificate location", e);
```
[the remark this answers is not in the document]
TimurSadykov: added

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +439,170 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
+          keyCacheBuilder.put(keyId, buildPublicKey(publicKeyPem));
+        }
+      } else {
+        for (JsonWebKey key : jwks.keys) {
+          try {
+            keyCacheBuilder.put(key.kid, buildPublicKey(key));
+          } catch (NoSuchAlgorithmException
+              | InvalidKeySpecException
+              | InvalidParameterSpecException ignored) {
+            ignored.printStackTrace();
```
suztomo, line 534: Can you use logger? I see `java.util.logging.Logger`? I see `log(Level level, String msg, Throwable thrown)` is used in the Credential class in this repository.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +439,170 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
+      parameters.init(new ECGenParameterSpec("secp256r1"));
+      ECParameterSpec ecParameters = parameters.getParameterSpec(ECParameterSpec.class);
+      ECPublicKeySpec pubSpec = new ECPublicKeySpec(pubPoint, ecParameters);
+      KeyFactory kf = KeyFactory.getInstance("EC");
+      return kf.generatePublic(pubSpec);
+    }
+  }
+
+  /** Custom exception for wrapping all verification errors. */
+  public static class VerificationException extends Exception {
```
suztomo, line 592: Can you check whether this needs to be public? Given `verify` catches it without exposing it to users, package-private seems appropriate to me.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +439,170 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
+    }
+
+    private PublicKey buildPublicKey(String publicPem)
+        throws CertificateException, UnsupportedEncodingException {
+      return CertificateFactory.getInstance("X.509")
+          .generateCertificate(new ByteArrayInputStream(publicPem.getBytes("UTF-8")))
+          .getPublicKey();
+    }
+
+    private PublicKey buildRs256PublicKey(JsonWebKey key)
```
suztomo, line 560: BuildRs256PublicKey and buildEs256PublicKey are the same as the reference implementation. https://github.com/googleapis/google-auth-library-java/blob/1434839e704621dc21f902c81cb67e66f7ef1a05/oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java#L369 Good.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -142,9 +204,87 @@ public final String getIssuer() {
...
+    }
+
+    // This method validates token signature per current OpenID Connect Spec:
+    // https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
+    // By default, method gets a certificate from well-known location
+    // A request to certificate location is performed using
+    // {@link com.google.api.client.http.javanet.NetHttpTransport}
+    // Both certificate location and transport implementation can be overridden via {@link Builder}
+    // not recommended: this check can be disabled with OAUTH_CLIENT_SKIP_SIGNATURE
+    // environment variable set to true.
```
suztomo, line 223: This seems repeating the Javadoc of `verifySignature`.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -142,9 +204,87 @@ public final String getIssuer() {
...
+    try {
+      return verifySignature(idToken);
+    } catch (VerificationException ex) {
```
suztomo, line 226: Logging? All these good messages in the exception are ignored here.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -142,9 +204,87 @@ public final String getIssuer() {
    * @return {@code true} if verified successfully or {@code false} if failed
    */
   public boolean verify(IdToken idToken) {
-    return (issuers == null || idToken.verifyIssuer(issuers))
-        && (audience == null || idToken.verifyAudience(audience))
-        && idToken.verifyTime(clock.currentTimeMillis(), acceptableTimeSkewSeconds);
+    boolean simpleChecks =
```
suztomo, line 207: `simpleChecks` and the subsequent `if (!simpleChecks)` are difficult to read for me. Do you have a better variable name? `IssuerAudienceAndTokenTimeAreValid` might be too long.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -142,9 +204,87 @@ public final String getIssuer() {
...
+      return verifySignature(idToken);
+    } catch (VerificationException ex) {
+      return false;
+    }
+  }
+
+  /**
+   * Verifies the signature part of the id token By default, method gets a certificate from
+   * well-known location A request to certificate location is performed using {@link
+   * com.google.api.client.http.javanet.NetHttpTransport} Both default can be overridden via {@link
```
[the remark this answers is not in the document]
suztomo: "Both defaults"? English is difficult.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: yes, probably makes sense

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: this is an entry point, more important. Since we don't have other ways to communicate i would rather over explain

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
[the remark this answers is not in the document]
TimurSadykov: yep )

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -142,9 +204,87 @@ public final String getIssuer() {
    * @return {@code true} if verified successfully or {@code false} if failed
    */
   public boolean verify(IdToken idToken) {
```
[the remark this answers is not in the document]
TimurSadykov: ah ok, will move that doc to here

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: fixed

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: renamed to tokenFieldsValid

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: Done, except current interface does not let us hav exception message, so it goes to docs and logger

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
suztomo: That's clear. Thanks.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +455,174 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
+            io);
+        return ImmutableMap.of();
+      }
+
+      ImmutableMap.Builder<String, PublicKey> keyCacheBuilder = new ImmutableMap.Builder<>();
+      if (jwks.keys == null) {
+        // Fall back to x509 formatted specification
+        for (String keyId : jwks.keySet()) {
+          String publicKeyPem = (String) jwks.get(keyId);
+          keyCacheBuilder.put(keyId, buildPublicKey(publicKeyPem));
```
suztomo, line 545: I believe `buildPublicKey` returning null is unexpected shouldn't happen. Would you add logging or source code comment for such null cases?

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +455,174 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
+        }
+      } else {
+        for (JsonWebKey key : jwks.keys) {
+          try {
+            keyCacheBuilder.put(key.kid, buildPublicKey(key));
```
suztomo, line 550: I believe `buildPublicKey` returning null would be unexpected. But, for the future writer or troubleshooter, would you add logging for such null case?

Or a source code comment explaining that buildPublicKey doesn't return null would be helpful. (I read that the null case shouldn't happen happen because the algorithm field already checked for the valid values.)

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -216,6 +379,18 @@ public Builder setIssuer(String issuer) {
       }
     }
 
+    /**
+     * Override the location URL that contains published public keys. Defaults to well-known Google
```
suztomo, line 383:
```suggestion
     * Overrides the location URL that contains published public keys. Defaults to well-known Google
```

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -280,5 +455,174 @@ public Builder setAcceptableTimeSkewSeconds(long acceptableTimeSkewSeconds) {
...
+    }
+
+    /** Sets the environment. Used mostly for testing */
+    Builder setEnvironment(Environment environment) {
+      this.environment = environment;
+      return this;
+    }
+
+    /**
+     * Set the HttpTransportFactory used for requesting public keys from the certificate URL. Used
```
suztomo, line 471:
```suggestion
     * Sets the HttpTransportFactory used for requesting public keys from the certificate URL. Used
```

suztomo (approved):

suztomo:
@chingor13 PTAL

suztomo:
Memo: OwlBot reverted the lint fix (#863) in the commit below (right after I merged 863), and now lint is failing in main (So this PR 861 is irrelevant to the lint failure). The lint is not a required check.

https://github.com/googleapis/google-oauth-java-client/pull/864/commits/b2b2e43cc84c842d52376a7b4a6c7206bc1e6d4c

chingor13 (approved):

EmpireMyDev (commented):
Nice

PR #892, opened 2022-05-31
fix: fix IdTokenVerifier so it does not cache empty entries

### LINKED ISSUE #891: Failure to refresh public keys causes IdTokenVerifier to fail valid tokens
After rolling out the v1.33.3 release, we've experienced a few transient 1 hour outages for OIDC token verification.

Logs show:

```
message: WARNING: Failed to get a certificate from certificate location https://www.googleapis.com/oauth2/v3/certs
```

And then all attempts to validate tokens fail with:
```
com.google.api.client.auth.openidconnect.IdTokenVerifier$VerificationException: Could not find PublicKey for provided keyId: 486f16482005a2cdaf26d9214018d029ca46fb56
```

For 1 hour straight until the cache of the empty map expires and a new cert refresh succeeds:
https://github.com/googleapis/google-oauth-java-client/blob/c1b1468807b97472665c0e218ffcf6955ec57190/google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java#L165

We believe that the implementation should not evict previously known public keys in the event of a refresh failure (which could happen for many different reasons).

TimurSadykov:
Fixes idtoken verifier so that it will not cache empty entries in cases of intermittent problems accessing the keystore

Also, it makes the payload verification protected so child classes can use it separately in case they do signature verification themselves

Fixes #891 ☕️


google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -534,7 +560,7 @@ public Map<String, PublicKey> load(String certificateUrl) throws Exception {
             Level.WARNING,
             "Failed to get a certificate from certificate location " + certificateUrl,
             io);
-        return ImmutableMap.of();
+        throw io;
```
suztomo, line 563: Memo for myself.

Now this throws an exception. With Guava's LoadingCache, the exception surfaces as ExecutionException in `publicKeyCache.get` in `verifySignature` method. This `verifySignature` method wraps it as `VerificationException`. The `verify` method catches `VerificationException`, logs the SEVERE-level message, and returns false.

With this change the LoadingCache now stops caching failure result. A potential problem would be the library users calling the verify method repeatedly in a situation where the certificate location is inaccessible. From the caller's perspective, the result (getting `false`) remains the same. In that case `IdTokenVerifier$PublicKeyLoader` logs the warning message. The user can easily tell the problem. This should be fine.

suztomo (approved):

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: Yep, had similar thoughts on this

PR #923, opened 2022-07-22
feat: Add googleapis-auth as codeowner team

lqiu96:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-oauth-java-client/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> ☕️

If you write sample code, please follow the [samples format](
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md).

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -61,3 +61,5 @@ permissionRules:
     permission: admin
   - team: yoshi-java
     permission: push
+  - team: googleapis-auth
+    permission: push
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,7 +4,9 @@
 # For syntax help see:
 # https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
-*                       @googleapis/yoshi-java
+# The @googleapis/googleapis-auth is the default owner for changes in this repo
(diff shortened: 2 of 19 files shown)


.github/sync-repo-settings.yaml
```diff
@@ -61,3 +61,5 @@ permissionRules:
     permission: admin
   - team: yoshi-java
     permission: push
+  - team: googleapis-auth
```
blakeli0, line 64: I think this change might be overridden to this [file](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/java_library/.github/sync-repo-settings.yaml) by the nightly owlbot template sync job, so we may need to add this file to the ignore list in `owlbot.py`. Is that correct @suztomo ?

suztomo: I just added Write role to googleapis-auth team via 'Settings'. Edit to .github/sync-repo-settings.yaml is not needed.

I observed OwlBot does not override the file https://github.com/googleapis/synthtool/pull/1494#discussion_r923910542

suztomo (approved):

PR #934, opened 2022-08-19
fix!: IdTokenVerifier now throws IOException if any issue obtaining public keys. Adding retries to public key fetch to cover transient network issues.

### LINKED ISSUE #916: Fix IdTokenVerifier interface by adding Exceptions in case of signature verification issues
Currently IDTokenVerifier.verify method can return false-negatives in case of issues with signature verification.
We want to fix the method signature to throw an exception and provide a better way of handling errors.

TimurSadykov:
IdTokenVerifier now throws IOException if any issue obtaining public keys. 

Adding 2 retries to public key fetch to cover transient network issues.

Fixes #916  ☕️


google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -242,11 +247,7 @@ public boolean verify(IdToken idToken) {
     try {
       return verifySignature(idToken);
     } catch (VerificationException ex) {
-      LOGGER.log(
-          Level.SEVERE,
-          "id token signature verification failed. "
-              + "Please see docs for IdTokenVerifier for default settings and configuration options",
-          ex);
+      LOGGER.log(Level.INFO, "Id token signature verification failed. ", ex);
```
sai-sunder-s, line 250: shouldn't log level be higher?

sai-sunder-s (approved):

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: Previously it was higher because we swallowed the potential public key fetch error problem... 

Now it is part of normal operation.

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -231,8 +234,10 @@ public final Collection<String> getAudience() {
    *
    * @param idToken ID token
    * @return {@code true} if verified successfully or {@code false} if failed
+   * @throws IOException thrown if verification failed to run. For example, failure to get public
```
alicejli, line 237: nit: consider editing to say `@throws IOException if verification failed to run.`

google-oauth-client/src/test/java/com/google/api/client/auth/openidconnect/IdTokenVerifierTest.java
```diff
@@ -318,15 +320,23 @@ public void testVerifyRs256TokenWithLegacyCertificateUrlFormat()
...
+    MockClock clock = new MockClock(1660880973000L);
     IdTokenVerifier tokenVerifier =
         new IdTokenVerifier.Builder()
             .setClock(clock)
             .setCertificatesLocation(SERVICE_ACCOUNT_CERT_URL)
             .setHttpTransportFactory(new DefaultHttpTransportFactory())
             .build();
     assertTrue(tokenVerifier.verify(IdToken.parse(JSON_FACTORY, SERVICE_ACCOUNT_RS256_TOKEN)));
+
+    // a token with bad signature expected to fail in verify and work in verifyPayload
```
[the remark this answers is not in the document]
TimurSadykov: fixed, thanks )

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -231,8 +234,10 @@ public final Collection<String> getAudience() {
...
+   *     keys for signature validation.
```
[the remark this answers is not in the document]
TimurSadykov: thanks, updated )

PR #960, opened 2022-11-02
fix: disabling integration test for idTokenVerifier until the infra is updated

TimurSadykov:
The verifier test works, however, the SA with the original token for verification was deleted. We need to create a new one. 

This is effectively an integration test, and the verifier is not required for any auth flow. It is an optional tool to validate IdToen if they are moved across VMs.


blakeli0 (approved):

TimurSadykov:
not needed

PR #1055, opened 2023-06-05
test: fix signature test

chingor13:
Use the mock http transport to stub the public key fetching from https://www.googleapis.com/oauth2/v3/certs. Traditionally, these public keys have not rotated (often). For the unit test, we lock the time (for timestamp expiry) and the existing public keys at this point in time.


chingor13:
Note: will fix the lint in a follow-up PR.

suztomo (approved):
unit test passed

TimurSadykov (approved):
LGTM, many thanks )

PR #1056, opened 2023-06-06
chore: fix code formatting

chingor13:
Updates the code formatter plugin to match what's in the google-cloud-shared-config. The post-processor seems to automatically run the code formatter, but the version needs to match or the test will not pass.

Related #1058


TimurSadykov (approved):
LGTM

PR #1082, opened 2023-11-20
chore: Update release-please versioning to "always-bump-minor"

ddixit14:
This versioning strategy always only bumps the minor version irrespective of the changes committed.  We can remove this after the minor-version release and releases will fall back to default versioning strategy. 
Reference: https://github.com/googleapis/release-please/blob/main/docs/customizing.md#versioning-strategies

### DIFF
--- a/.github/release-please.yml
+++ b/.github/release-please.yml
@@ -1,6 +1,7 @@
 bumpMinorPreMajor: true
 handleGHRelease: true
 releaseType: java-yoshi
+versioning: always-bump-minor
 branches:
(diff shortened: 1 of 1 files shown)


ddixit14:
`release-please config schema` is failing but it should be okay. Raised a [PR](https://github.com/googleapis/repo-automation-bots/pull/5262) to fix it in release-please source code. The versioning strategy should work irrespective of this schema failure.

TimurSadykov (approved):
LGTM, Thanks @ddixit14

PR #1091, opened 2023-12-07
fix: deprecating the IdTokenVerifier.verify, adding verifyOrThrow as an alternative

TimurSadykov:
This PR restores the interface of the IdTokenVerifier.verify method to avoid a previous breaking [change](https://github.com/googleapis/google-oauth-java-client/pull/934/files#diff-614b71029b58782f8684645ff56c443eac1ab67b0f6736911ae100d7e3971f8f) that was not released yet. It also deprecates the method as it is no longer recommended to use because it can produce false-negatives if public keys for signature verification are not available. 

Here we add the IdTokenVerifier.verifyOrThrow as an alternative that throws IOExceptions if any issues getting public keys.

### DIFF
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
@@ -227,9 +227,9 @@ public class IdTokenVerifier {
    *       variable set to true. Use {@link #verifyPayload(IdToken)} instead.
    * </ul>
    *
-   * Deprecated. This method returns false if network requests to get certificates fail. Use
-   * {@link IdTokenVerifier.verfyOrThrow(IdToken)} instead to differentiate between potentially
-   * retryable network errors and false verification results.
+   * Deprecated. This method returns false if network requests to get certificates fail. Use {@link
+   * IdTokenVerifier.verfyOrThrow(IdToken)} instead to differentiate between potentially retryable
+   * network errors and false verification results.
    *
    * @param idToken ID token
    * @return {@code true} if verified successfully or {@code false} if failed
--- a/google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
+++ b/google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
@@ -25,7 +25,6 @@ import com.google.api.client.json.GenericJson;
 import com.google.api.client.json.gson.GsonFactory;
 import com.google.api.client.json.webtoken.JsonWebSignature.Header;
 import com.google.api.client.util.Base64;
-import com.google.api.client.util.Beta;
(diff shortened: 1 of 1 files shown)


arithmetic1728 (approved):

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -230,14 +227,52 @@ public final Collection<String> getAudience() {
...
+   *       IdToken#verifyAudience(Collection)}.
+   *   <li>The current time against the issued at and expiration time, using the {@link #getClock()}
+   *       and allowing for a time skew specified in {@link #getAcceptableTimeSkewSeconds()} , by
+   *       calling {@link IdToken#verifyTime(long, long)}.
+   *   <li>This method verifies token signature per current OpenID Connect Spec:
+   *       https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation. By default,
+   *       method gets a certificate from well-known location. A request to certificate location is
+   *       performed using {@link com.google.api.client.http.javanet.NetHttpTransport} Both
+   *       certificate location and transport implementation can be overridden via {@link Builder}
+   *       not recommended: this check can be disabled with OAUTH_CLIENT_SKIP_SIGNATURE environment
```
westarle, line 264: Nit: if it isn't recommended we might want to remove the capability from our new API.  WDYT?

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
```diff
@@ -230,14 +227,52 @@ public final Collection<String> getAudience() {
    *       variable set to true.
    * </ul>
    *
-   * <p>Overriding is allowed, but it must call the super implementation.
+   * Deprecated, because returns a false-negatives in case of an error while getting public keys for
```
westarle, line 230: Nit: I would say "This method returns false if network requests to get certificates fail.  Use verifyOrThrow instead to differentiate between potentially retryable network errors and false verification results."

westarle (approved):

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: Applied

TimurSadykov: Thanks!

google-oauth-client/src/main/java/com/google/api/client/auth/openidconnect/IdTokenVerifier.java
TimurSadykov: As discussed offline - leaving it here since this verifier historically allowed only payload verification and we have same env variable in the newer library

PR #1093, opened 2023-12-14
Add ProGuard rules for this library

Goooler:
Add ProGuard/R8 keep rules to ensure this lib will be compatible with obfuscation enabled, the files under `src/main/resources/META-INF/proguard` should be interpreted by R8 automatically.

- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-oauth-java-client/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Appropriate docs were updated (if necessary)


Goooler:
Actually, we can just keep a few classes under `com.google.api.client.auth` instead of all of them, but I have no more time to investigate that.

meltsufin (approved):

PR #1135, opened 2025-01-09
ci: Airlock-friendly build dependencies

suztomo:
Airlock needs newer version of the Maven plugins. We also do not need to declare nexus-staging-maven-plugin by default.


meltsufin (approved):
Should we use shared config as parent?

suztomo:
I tried that idea and in theory it should work but it became messy.

PR #1145, opened 2025-01-30
fix: animal-sniffer and clirr to a profiles

suztomo:
b/384085175#comment40

We can skip animal-sniffer-maven-plugin and CLIRR plugin execution if we wrap it as a profile.

### DIFF
--- a/google-oauth-client-appengine/pom.xml
+++ b/google-oauth-client-appengine/pom.xml
@@ -62,18 +62,6 @@
           </execution>
         </executions>
       </plugin>
-      <!--App Engine uses Java 7 or Java 8-->
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>animal-sniffer-maven-plugin</artifactId>
-        <configuration>
-          <signature>
-            <groupId>org.codehaus.mojo.signature</groupId>
-            <artifactId>java17</artifactId>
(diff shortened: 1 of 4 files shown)


google-oauth-client-appengine/pom.xml
```diff
@@ -62,18 +62,6 @@
           </execution>
         </executions>
       </plugin>
-      <!--App Engine uses Java 7 or Java 8-->
-      <plugin>
```
suztomo, line 66: The parent already defines the plugin declaration.

google-oauth-client-java6/pom.xml
```diff
@@ -47,18 +47,6 @@
           </execution>
         </executions>
       </plugin>
-      <!--Set minimum version to Java 7-->
-      <plugin>
```
suztomo, line 51: The parent already defines the plugin declaration.

google-oauth-client-jetty/pom.xml
```diff
@@ -49,18 +49,6 @@
           </execution>
         </executions>
       </plugin>
-      <!--Minimum version is Java 7 -->
-      <plugin>
```
suztomo, line 53: The parent already defines the plugin declaration.

meltsufin (approved):

